PR #147, opened 2015-01-14
Add unix-timestamp-to-timestamptz transform

erjiang:
This patch was useful to us in loading some data from one database to another, and I thought it may be useful to others as well. I appreciate that it was easy to add custom data transform functions to pgloader - thanks for a very useful tool!

### DIFF
--- a/src/utils/transforms.lisp
+++ b/src/utils/transforms.lisp
@@ -227,3 +227,20 @@
 
                    (t
                     date-string-or-integer)))))))
+
+(defun unix-timestamp-to-timestamptz (unixtime-string)
+  "Takes a unix timestamp (seconds since beginning of 1970) and converts it
+  into a string of format \"YYYY-MM-DD hh:mm:ssZ\".
+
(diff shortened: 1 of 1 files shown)


dimitri:
Thanks for your contribution!

I'm a little worried about putting the computed constant as a literal in the code, could you rework your patch using http://lisptips.com/post/11649360174/the-common-lisp-and-unix-epochs for inspiration on how to get it in a more dynamic way?

Rather than add top-level bindings with `defvar`, you could also use `load-time-value` in your `let` bindings, see http://www.lispworks.com/documentation/HyperSpec/Body/s_ld_tim.htm

Once done consider rebasing your PR (squash it into a single commit) so that I can merge your code.

erjiang:
I'm not sure I share your worry about the constant, but I can see that it's easier to verify the code if it's computed at load time.

Anyways, the PR has been updated.

dimitri:
Thanks for the update, I prefer it that way indeed. Merging!

PR #151, opened 2015-01-18
add missing conditional compilation for sbcl

zickzackv:
Added missing #+sbcl for sb-ext:*muffled-warnings*.  prevented succesfully building the application with ccl


dimitri:
Thanks!

PR #152, opened 2015-01-19
add note to install universal binaries of freetds

zickzackv:
Otherwise cffi tries to load the 64-bit version of the library which fails


dimitri:
Thanks!

PR #165, opened 2015-01-30
Missing quotes/escape on table names

achamo:
Hello,

If you try to import a database with tables named "user" or "group", it won't work.
I add some quotes on sqlite files, I didn't check other sources :)


dimitri:
Please see about using `apply-identifier-case` as in the MySQL case.

dimitri:
Closing, the source code changed a lot in between, and the current tests include a `"user"` table in a SQLite database, so we should be covered.

Issue #168, opened 2015-02-04
Can not connect to MySQL if there is a dot in the user name


pinaraf:
Hi

I'm trying to connect as user firstname.lastname to a mysql database, but pgloader does not recognize the mysql connection string and try to parse it as a sqlite file path :

   Failed to open sqlite file #P"/home/pinaraf/mysql:/firstname.lastname:XXYYXXYY@1.2.3.4:5678 paiement.load": Could not open sqlite3 database /home/pinaraf/mysql:/firstname.lastname:XXYYXXYY@1.2.3.4:5678/paiement.load

It seems to be a regression from previous older pgloader3 releases where this was not crashing.
I've fixed the issue, pushing a merge request in a few secs…

dimitri:
Closing, should be good now.

PR #169, opened 2015-02-04
Allow dots in user names

pinaraf:
Fix issue #168 by adding . in the punct defrule.

### DIFF
--- a/src/parsers/command-db-uri.lisp
+++ b/src/parsers/command-db-uri.lisp
@@ -25,7 +25,14 @@
 (defrule doubled-at-sign (and "@@") (:constant "@"))
 (defrule doubled-colon   (and "::") (:constant ":"))
 (defrule password (+ (or (not "@") doubled-at-sign)) (:text t))
-(defrule username (and namestring (? (or doubled-at-sign doubled-colon)))
+(defrule username-string (and (or #\_ (alpha-char-p character))
+                              (* (or (alpha-char-p character)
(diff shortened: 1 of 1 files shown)


dimitri:
I don't think your fix can be used at all really, because the `punct` rule is used in the definition of a `namestring` and only there. The fix I envision would add a new `mysql-user-password` rule to use in the `mysql-uri` rule, wherein it's using a `mysql-username` rule that allows dots in names.

Do you want to see about that?

pinaraf:
Well, my «fix» was enough to make it work for my need, but I understand it may be a terrible way to fix. It just seemed weird to me that a punct role did not include a dot… Well, must be a translation misunderstanding.
namestring is the rule used in a lot of rules, and is also used in the mysql user name rule.
I've modified my branch, I've not used the "mysql-username" name for the rule because that same rule is used for postgresql user name (that suffered from the same bug…)

dimitri:
I'm not sure dots are allowed in PostgreSQL user names, will look into that. The problem with editing the  *namestring* rule is that it's then used for table names where dots are not allowed, as they are used as a separator in between a schema name and a table name.

pinaraf:
Dots are allowed in PostgreSQL user names, we use them in production here.

dimitri:
So, getting back to you, I see that your fix is actually correct, when I though you were still touching the *namestring* rule itself. Applying, thanks!

PR #181, opened 2015-02-19
Fix issue 180: proper quoting of already quoted identifiers.

victorkryukov:


victorkryukov:
Thanks for the fix, and sorry for making one commit on top of another.

PR #186, opened 2015-02-25
Fix 182: properly quote tables with . in their names

### LINKED ISSUE #182: MS SQL Tables with "." in their name are not imported properly
Trying to migrate my (very hairy) database, I've encounted the following problem: pgloader dies after a while with the following message:
```
An unhandled error condition has been signalled:
   Postgresql protocol error: Unexpected message received: Z
```
Looking at the `pgloader.log`, the last statement is
```
2015-02-18T17:20:13.086000-08:00 NOTICE COPY dbo.dbo.opt_task_report
```
I have two tables in my database: `opt_task_report` and `dbo.Opt_Task_Report`, both belonging to `dbo` schema. The latter was clearly mistakenly created at some point, and was never used.

However, `pgloader` hasn't recreated the later table schema (it does seem to recreate all other tables at this point), and doesn't seem to quote the identifier properly, and since you cannot use the name directly without quoting:
```
test-cs=# select * from dbo.dbo.opt_task_report;
ERROR:  cross-database references are not implemented: "dbo.dbo.opt_task_report"
LINE 1: select * from dbo.dbo.opt_task_report;
                      ^
Time: 0.170 ms
```
I think issueing a `COPY` statement with unquoted table name results in a similar error messages which throws `pgloader` off.

Of course, I can just remove (or rename) the table in question, but maybe it's another opportunity to make MS SQL import more robust. Looking forward for your suggestions.

vkryukov:
Seems like it was a typo. Fixed #182 for me.

### DIFF
--- a/src/pgsql/schema.lisp
+++ b/src/pgsql/schema.lisp
@@ -26,7 +26,7 @@
           (cond ((quoted-p identifier)
                  :none)
 
-                ((cl-ppcre:scan "[^A-Za-z_]" identifier)
+                ((not (cl-ppcre:scan "^[A-Za-z_][A-Za-z0-9_$]*$" identifier))
                  :quote)
 
                 ((member lowercase-identifier *pgsql-reserved-keywords*
--- a/src/pgsql/schema.lisp
+++ b/src/pgsql/schema.lisp
@@ -26,7 +26,7 @@
           (cond ((quoted-p identifier)
                  :none)
 
-                ((cl-ppcre:scan "^[^A-Za-z_]" identifier)
+                ((cl-ppcre:scan "[^A-Za-z_]" identifier)
                  :quote)
 
                 ((member lowercase-identifier *pgsql-reserved-keywords*


victorkryukov:
```
$ pgloader opt.txt 
2015-02-24T16:51:09.052000-08:00 LOG Main logs in '/private/tmp/pgloader/pgloader.log'
2015-02-24T16:51:09.054000-08:00 LOG Data errors in '/private/tmp/pgloader/'
2015-02-24T16:51:09.054000-08:00 LOG Parsing commands from file #P"/Users/victor/src/pgloader/opt.txt"
Max connections reached, increase value of TDS_MAX_CONN
2015-02-24T16:51:09.259000-08:00 WARNING Postgres warning: table "dbo.Opt_Task_Report" does not exist, skipping
Max connections reached, increase value of TDS_MAX_CONN
                    table name       read   imported     errors            time
------------------------------  ---------  ---------  ---------  --------------
                   before load          1          1          0          0.013s
               fetch meta data          1          1          0          0.075s
              create, truncate          0          0          0          0.021s
------------------------------  ---------  ---------  ---------  --------------
     dbo."dbo.Opt_Task_Report"     168605     168605          0         10.841s
        Index Build Completion          0          0          0          0.000s
------------------------------  ---------  ---------  ---------  --------------
                Create Indexes          0          0          0          0.000s
               Reset Sequences          0          0          0          0.021s
                  Primary Keys          0          0          0          0.000s
                  Foreign Keys          0          0          0          0.000s
------------------------------  ---------  ---------  ---------  --------------
             Total import time     168605     168605          0         10.971s
```

dimitri:
Well it was short of a brick load, but not a typo. It's implementing (or trying to) http://www.postgresql.org/docs/9.1/static/sql-syntax-lexical.html wherein

> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable.

Do you mind enhancing the regexp test to match that description better?

victorkryukov:
Yep - I just noticed I'm not passing some tests. Let me have a look.

victorkryukov:
@dimitri OK, I updated the regexp, but I suspect the test was failing for a different reason. Could you please have a look?

dimitri:
Looks good to me, and I'm quite convinced your dotted table name problem comes from here. Which test was failing? As soon as travis confirms the build I'm going to merge ;-)

dimitri:
Build failed. Too late in the night for me to see about it, will look into it tomorrow.

victorkryukov:
The one with `csv.blocks` - notice there is a dot in the name, since it creates a schema for some reason.

We could also make this regexp understand Unicode and all proper letter out there if we load `cl-ppcre-unicode` package, but my ASDF/package fu is not strong enough to do this myself. I will update the regexp though, but it will fail unless this package is loaded.
```lisp
* (cl-ppcre:scan "^[\\p{letter}_][\\p{letter}0-9$_]*$" "ФормальностьTh12341234$$$$")

0
26
#()
#()
```

victorkryukov:
Ok, I think I have an idea about what's causing the problem, but have no idea how to fix it - I would need to understand the code base much better in order to do this, so that would need to wait for another couple of weeks until I have more free time:

Basically, after we started quoting dots, the following doesn't work as intended - I think we are quoting everything after `?` in `INTO`, and thus trying to insert into table `"csv.blocks"` instead of table `blocks` of schema `csv`, but this table obviously doesn't exist.

Looks like parse would need to make an exception about allowing one, but not two, dots in the name before quoting for `INTO postgresql:///xxx/?table.name` syntax.

```
LOAD CSV
    FROM INLINE
         WITH ENCODING iso-8859-1
         (
            startIpNum, endIpNum, locId
         )
    INTO postgresql:///pgloader?csv.blocks
         (
            iprange ip4r using (ip-range startIpNum endIpNum),
            locId
         )
    WITH truncate,
         fields optionally enclosed by '"',
         fields escaped by double-quote,
         fields terminated by ','
```

victorkryukov:
Thinking about it more, I see no easy fix. We are passing two kinds of names around:

- The ones that are received from untrusted source (such as when querying a foreign db schema) and thus need to be quoted
- The ones that are coming from trusted source, such as from a pgloader command file, where use may have used a dot in the table name to denote a schema, and need NOT to be quoted.

However, e.g. `truncate-tables` use `apply-identifier-case` indiscriminately. 

Maybe the solution is obvious to you as you actually know the codebase :).

dimitri:
I just pushed a fix to the situation where you could not make heads or tails of the table name containing being qualified or not. Now as you can see qualified table names are given as a `cons` from the parser whereas are always strings when given by another source.

Please see again about your patch, I think now you should be able to make it work correctly.

victorkryukov:
Thank you @dimitri! Will have a look.

dimitri:
After some more fixes, I could actually cherry-pick your patches in, hoping that it will help fix #287 too. Thanks again, and sorry it took that much time for me to clean-up before being able to apply your work!

PR #191, opened 2015-03-09
Fix incorrect dates in specfile

mtyson01:
RPM complains about these dates being invalid during the build process.

I'm not sure which date exactly you mean, but I've changed the name of the day to match the number so the RPM can be built without errors.


dimitri:
Ahah, thanks ;-)

PR #196, opened 2015-03-23
Match pgloader-standalone target to pgloader target

benesch:
It may have taken me a year, but [pgloader is finally in Homebrew](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/pgloader.rb)!

This PR brings a couple updates to the `pgloader-standalone` target to bring it in line with updates to the `pgloader` target. Things seem to work without this update, but I didn't test the SSL mess that `hooks.lisp` is supposed to address. (I know this dual-target Makefile is far from ideal, but I can't think of a better solution.)


dimitri:
Hey, thanks a lot for that! ;-)

PR #201, opened 2015-04-02
Fix typo

jdufresne:


dimitri:
The example line 939 (fixed file format) should not have its trailing whitespace removed, see https://github.com/dimitri/pgloader/blob/master/test/fixed.load. Other than that, thanks for the patch, once fixed I'll merge it!

jdufresne:
> The example line 939 (fixed file format) should not have its trailing whitespace removed

Oops. Sorry about that. Fix in the latest version.

dimitri:
Thanks!

Issue #207, opened 2015-04-10
MSSQL "((0))" "(0)" default values not working


luqelinux:
Hi,

is it possible to workaround this?

```
An unhandled error condition has been signalled:
   The value "((0))" is not of type (OR NULL SINGLE-FLOAT DOUBLE-FLOAT).


Date/time: 2015-04-10-22:36An unhandled error condition has been signalled:
                              The value "((0))"
                              is not of type
                                (OR NULL SINGLE-FLOAT DOUBLE-FLOAT).


Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {10073EE903}>
0: (PGLOADER.TRANSFORMS::FLOAT-TO-STRING "((0))") [tl,external]
1: (PGLOADER.SOURCES::FORMAT-PGSQL-DEFAULT-VALUE "((0))" PGLOADER.TRANSFORMS::FLOAT-TO-STRING)
2: (PGLOADER.SOURCES::FORMAT-PGSQL-TYPE (:TABLE-NAME "x" :COLUMN-NAME "y" :TYPE "decimal" :CTYPE "decimal(18,2)" :TYPEMOD (18 . 2) :DEFAULT "((0))" :NOT-NULL T :AUTO-INCREMENT NIL) (:TYPE "numeric") PGLOADER.TRANSFORMS::FLOAT-TO-STRING)
```

dimitri:
This reminds me of the work we did at https://github.com/dimitri/pgloader/commit/7ccbb45763d17a99a252e400f98d1a6dec3dcb89 from the issue #162. I still would like to understand in which case does MS SQL register `((0))` as a default value and if there's a sane way not to get that from our catalog queries.

Here's the query we use to get the columns meta-data, maybe you know of a way to get a cleaner/easier value to deal with than what `information_schema.columns.column_default` outputs?

```sql
  select c.table_schema,
         c.table_name,
         c.column_name,
         c.data_type,
         c.column_default,
         c.is_nullable,
         COLUMNPROPERTY(object_id(c.table_name), c.column_name, 'IsIdentity'),
         c.CHARACTER_MAXIMUM_LENGTH,
         c.NUMERIC_PRECISION,
         c.NUMERIC_PRECISION_RADIX,
         c.NUMERIC_SCALE,
         c.DATETIME_PRECISION,
         c.CHARACTER_SET_NAME,
         c.COLLATION_NAME

    from information_schema.columns c
         join information_schema.tables t
[...]

PR #210, opened 2015-04-17
Fix for #207

luqelinux:
Not sure why "0" is not interpreted as integer, string type used instead.


dimitri:
That's because the SQL query you write returns it as a string, and CL follows on what MS SQL query meta data say: it's a string. Now, could you please rebase your Pull Request so that we have a single commit showing up with changes to both files?

dimitri:
I finally just pushed a fix to #207 mixing both your patch and an even more sophisticated expression for defaults as per @stans. Thanks for your efforts!

PR #213, opened 2015-04-23
Adds bootstrap for CentOS 7

gvangool:
This corrects the bootstrap for CentOS 7.

Notes:
- During installation you still need to do the following:

  ```sh
sudo yum -y install freetds
sudo ln -s /usr/lib64/libsybdb.so.5 /usr/lib64/libsybdb.so
```

  I'm not sure which package depends on `libsybdb.so` to fix that.
- `--compress-core` is not available in the `sbcl` from epel. So I build the RPM with `make COMPRESS_CORE=no rpm`


dimitri:
Thanks!

PR #233, opened 2015-05-14
Accept periods in CSV field names

malept:
Periods are allowed in PostgreSQL column names as well. See the test file for an example of using periods with a fixed width file definition.


dimitri:
The regression test is failing because you didn't add files in the proper place. You can run the test suite locally with `make test`; please add-in your new files (looks like you have forgotten them in your PR).

Also please have a look at https://github.com/dimitri/pgloader/issues/182 to see an analysis of something quite similar, if only to double check you're not missing something here. I will then do the same of course ;-)

malept:
Ah yes, I forgot to add the expected output file. It has been added.

Regarding #182, I believe that I'm OK on that end, since I'm quoting all of my field names in my command file. When this patch isn't applied, `pgloader` fails to parse the command at the fixed width definition.

dimitri:
Yeah you're good here because the parsing rule can not be used in a context where splitting schema name and table name may bring confusion, so it's ok to accept dots in there, we know we can only parse a column name anyway.

Can you please *rebase* your PR so that I have a single commit to merge? Thanks.

malept:
> Can you please rebase your PR so that I have a single commit to merge?

Done.

dimitri:
Thanks!

PR #240, opened 2015-05-26
Two bugfixes

alexbaretta:
* Fix identifier case in FOREIGN KEY constraints
* Fix string delimiter syntax in COMMENT statements

### DIFF
--- a/src/pgsql/queries.lisp
+++ b/src/pgsql/queries.lisp
@@ -207,7 +207,7 @@
       (when tables
         (pomo:execute
          (format nil "create temp table reloids(oid) as values ~{('~a'::regclass)~^,~}"
-                 tables)))
+                 (mapcar #'apply-identifier-case tables))))
 
       (handler-case
           (let ((sql (format nil "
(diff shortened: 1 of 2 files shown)


dimitri:
Hi, thanks for the fix!

I don' t understand what you are achieving in the 5b3fb6fd1106331afb4c5ed842e400831f206351 patch tho, can you expand on that? All I can see is weakening of the poor man's dollar quoting protection I put into place here...

alexbaretta:
Hey Dmitri,

Thanks for taking time to look into this. The rationale for 5b3fb6f is that dashes are not supported in $-delimiter strings.

main=# select $ABC-DEF$Hello, Dmitri!$ABC-DEF$;
ERROR:  syntax error at or near "$"
LINE 1: select $ABC-DEF$Hello, Dmitri!$ABC-DEF$;
               ^

dimitri:
I am confused, because I (of course) tested that feature. Now it could be that the test case was wrong and didn't use the code path :( Anyways, can you amend your patch here to use "_" rather than "-", please?

alexbaretta:
main=# select $ABC_DEF$Hello, Dmitri!$ABC_DEF$;
    ?column?    
----------------
 Hello, Dmitri!
(1 row)

I guess I can...

dimitri:
Thanks!

alexbaretta:
My pleasure!

PR #241, opened 2015-05-27
Fix DROP TABLE statements on tables with foreign keys

alexbaretta:


dimitri:
What if you're loading a subset of the tables and using `WITH DROP`? I bet in that case it's better to have errors than to mistakenly drop the whole database table set?

alexbaretta:
I think the current behavior defeats the purpose of `WITH DROP`. This option is supposed to allow the user to repeatedly run the same pgloader script without having to manually drop and recreate the database.

dimitri:
That's true. Please amend the docs to reflect the `CASCADE` usage and its consequences (existing local tables are going to disappear, you will *LOSE* data, etc) then, and I will merge. Thanks!

alexbaretta:
Dimitri, I updated the markdown doc, but I have no idea how to update pgloader.1. I assume it's groff, and I have no knowledge of it.

dimitri:
There's a `make docs` target, but don't worry, I am due to fix the toolchain here and the manpage has not been updated when its source has been changed lately. Will take care of that before next release.

Can you rebase your pull request to make it a single commit, and fix *please not* into *please note* while at it? Thanks!

alexbaretta:
Done.

dimitri:
And merged, thanks!

alexbaretta:
Awesome!

PR #253, opened 2015-07-03
MySQL bigint default convert's to NUMERIC(20)

gregorg:
MySQL numeric data types are often 'optimized' to unsigned, so by default we need to convert MySQL's bigint to PG NUMERIC(20)

### DIFF
--- a/src/sources/mysql/mysql-cast-rules.lisp
+++ b/src/sources/mysql/mysql-cast-rules.lisp
@@ -75,7 +75,7 @@
     (:source (:type "mediumint") :target (:type "integer"  :drop-typemod t))
     (:source (:type "integer")   :target (:type "integer"  :drop-typemod t))
     (:source (:type "float")     :target (:type "float"    :drop-typemod t))
-    (:source (:type "bigint")    :target (:type "bigint"   :drop-typemod t))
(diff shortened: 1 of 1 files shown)


dimitri:
I'd prefer that we have a way to do that only for `unsigned` data types, but on a quick read of the type casting machinery I just failed to find the hints where to do that. Will review later (next week).

Meanwhile `--cast 'type bigint to numeric(20)'` in your command line (or load command) should do the trick.

gregorg:
I failed too, that's why I've choose to cast to numeric by default. IMHO casting bigint to bigint must be tricked by --cast parameter, cause in this case the loader known what he is doing.

dimitri:
Given the performances implications, it will be a POLA violation to migrate over to `numeric` in all cases, so I want to detect `unsigned bigint` and only pick `numeric` by default in that case. Also it gives me the perfect opportunity to implement type casting debug messages (at :info level maybe). So you will have to wait for a bit before I get back to it and implement the whole thing, but we'll be much better off.

dimitri:
Currently there's no way to express casting rules against the *unsigned* part of the MySQL `column_type`, so I just guarded against the number of allowed digits (the typmod precision field).

PR #267, opened 2015-07-26
reverse-*-*: fix double function definition typos for cols and tables

dkochmanski:
Functions #'reverse-indexes-cols and #'reverse-fkeys-cols were defined
twice in same lambda block, which was probably typo. This renames
second definitions to #'reverse-indexes-tables and

Signed-off-by: Daniel Kochmański <daniel@turtleware.eu>

Please verify, if this is indeed a typo - I didn't try to analize code to understand, what it's doing.

### DIFF
--- a/src/sources/mssql/mssql-schema.lisp
+++ b/src/sources/mssql/mssql-schema.lisp
@@ -252,13 +252,13 @@ order by SchemaName,
                   (loop :for (name . index) :in list-of-indexes
                      :collect (cons name (reverse-index-cols index))))
 
-                (reverse-indexes-cols (list-of-tables)
+                (reverse-indexes-tables (list-of-tables)
                   (reverse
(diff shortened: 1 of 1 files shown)


dimitri:
I think we'll need a test case here to be able to commit your changes, which do look sane. I'm quite confused about what the current code is doing here :(

dkochmanski:
Me neither, but what's the point of defining two local function, where second definition shadows first one? Basically it's something like:

    (labels ((xxx () "1")
             (xxx () "2"))
      (xxx))

dimitri:
Exactly, the code is wrong, no doubt. My hesitation is how to validate that the fix is right, as I don't really have mssql test cases to validate it with... :(

dimitri:
Fixed in 56a89e9b53e74d51b99248e4aacd4cc4d5dc8154.

PR #304, opened 2015-10-26
MSSQL: handle nvarchar fields properly.

fiddlerwoaroof:
PGLoader was choking on the default values of nvarchar fields
So I added a transform to strip the N'...' delimiters around nvarchar
(unicode) fields and convert them to text automatically.

https://msdn.microsoft.com/en-us/library/ms179899.aspx
https://msdn.microsoft.com/en-us/library/ms186939.aspx
http://stackoverflow.com/questions/144283/what-is-the-difference-between-varchar-and-nvarchar

### DIFF
--- a/src/sources/mssql/mssql-cast-rules.lisp
+++ b/src/sources/mssql/mssql-cast-rules.lisp
@@ -8,7 +8,10 @@
   `((:source (:type "char")      :target (:type "text" :drop-typemod t))
     (:source (:type "nchar")     :target (:type "text" :drop-typemod t))
     (:source (:type "varchar")   :target (:type "text" :drop-typemod t))
-    (:source (:type "nvarchar")  :target (:type "text" :drop-typemod t))
+
+    (:source (:type "nvarchar")  :target (:type "text" :drop-typemod t) 
+             :using pgloader.transforms::sql-server-nvarchar-to-text)
+
     (:source (:type "xml")       :target (:type "text" :drop-typemod t))
 
     (:source (:type "bit") :target (:type "boolean")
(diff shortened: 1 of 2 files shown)


dimitri:
Thanks for the bug report and pull request! Sorry about the delay to get to review them, pgconf.eu was keeping me very busy last week... I have a couple of questions about your proposed patch.

First, easily solved. The following notation incurs a useless copy of the input string:

    (string= "N'" (subseq result 0 2))

It would better be rewritten as

    (string= "N'" result :end2 2)

Other than that, what if the user string actually is something like `N'foo'`?

fiddlerwoaroof:
So, on inspecting the results of importing my data, I discovered that my code rewrote null values to the string "NIL".  This commit fixes that as well as implements your suggested improvements to the call to string=.

As far as your other question goes, I'm not sure what all the edge cases are.  Perhaps it'd be better to accept the new transformation function but leave the default untransformed?  The only problem with that is that it will expose people to the bug that made me interested in the first place: if a `N'...'` field is present in the MSSQL database, pgloader generates invalid SQL for inserting the new value into the DB. The generated code is something like:

```(lang=sql)
INSERT INTO mytable VALUES (mycol) 'N'value'';
```

Which seems to be not only open to this kind of problem but to accidental SQL injection as well.

dimitri:
Is the format `N'...'` always used for `nvarchar`or controllable in some fashion? is there some SQL expression we can use in the `SELECT` list to control it? pgloader already has facilities to use something else than `SELECT column_name`, see function `get-column-list` at the end of the file https://github.com/dimitri/pgloader/blob/master/src/sources/mssql/mssql-schema.lisp

dimitri:

```
```

Closing as fixed in v4.

PR #375, opened 2016-03-24
update fixed-width example to work with current dataset

gorthx:
- old URL was 404
- make loc_name/LocationName field name consistent;  original example
  threw an error: "The variable PGLOADER.TRANSFORMS::LOC_NAME is
  unbound."

### DIFF
--- a/web/src/fixed.md
+++ b/web/src/fixed.md
@@ -16,9 +16,9 @@ You can find more files from them at the
 Here's our command:
 
     LOAD ARCHIVE
-       FROM http://www.census.gov/geo/maps-data/data/docs/gazetteer/places2k.zip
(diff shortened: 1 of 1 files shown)


dimitri:
The real test file was already updated and good, as seen at https://github.com/dimitri/pgloader/blob/master/test/census-places.load. It took me a while to figure out your patch applies to the docs, not the test!

Can you update your patch to match with the test file, where we keep "LocationName" as a test of quoting column names that need it?

gorthx:
Well, crumbs.  I haven't been able to get the example to work with "LocationName" on either my mac or my Ubuntu box.  I'll dig in to this a bit more & figure out what's going on.

Thanks!

PR #393, opened 2016-04-25
fix type drop to cascade

porshkevich:
if you have function or operator with  type which is removed, you will have error

error: cannot drop type because other objects depend on it


dimitri:
Thanks!

I was worried that the dependencies would remove too many objects, but the way we name the type makes it hard to experience a collision with other non-related user defined types...

PR #402, opened 2016-05-10
fix cast from float to real

porshkevich:
PostgreSQL support SQL-standard notations float and float(p) BUT without precision it select double precision instead real.

http://www.postgresql.org/docs/current/static/datatype-numeric.html
last paragraph of 8.1.3

### DIFF
--- a/src/sources/mysql/mysql-cast-rules.lisp
+++ b/src/sources/mysql/mysql-cast-rules.lisp
@@ -79,7 +79,7 @@
     (:source (:type "smallint")  :target (:type "smallint" :drop-typemod t))
     (:source (:type "mediumint") :target (:type "integer"  :drop-typemod t))
     (:source (:type "integer")   :target (:type "integer"  :drop-typemod t))
-    (:source (:type "float")     :target (:type "float"    :drop-typemod t))
+    (:source (:type "float")     :target (:type "real"    :drop-typemod t))
     (:source (:type "bigint")    :target (:type "bigint"   :drop-typemod t))
 
     (:source (:type "double")


dimitri:
Thanks for the patch! Do you have a link to the MySQL documentation where they say what their own `float` type is? You see that in the current cast rules, we're always targeting the `double precision` alternative. If we want to target either `real` or `float` then we will have to check the digit precision of the MySQL type too, which is not provided by your current patch.

porshkevich:
I reread.

Maybe I made a mistake somewhere.

Let me explain what I wanted to fix


MySQL | PG
---|---
float(without precision) | real
float(0-23) | real
float(24-53) | double precision
double | double precision

But in PG `float(without precision)` cast to `double precision`, different than in MySQL.

In PG `real` cast to internal `float4` (4-byte single-precision FLOAT) and `double precision` cast to `float8` (8-byte double-precision float)

```
postgres=# SELECT
postgres-#         '1.123456789123456789'::float,
postgres-#         '1.123456789123456789'::real,
postgres-#         '1.123456789123456789'::double precision,
postgres-#         '1.123456789123456789'::float(1),
postgres-#         '1.123456789123456789'::float(2),
postgres-#         '1.123456789123456789'::float(24),
postgres-#         '1.123456789123456789'::float(48);
-[ RECORD 1 ]------------
float8 | 1.12345678912346
float4 | 1.12346
float8 | 1.12345678912346
float4 | 1.12346
float4 | 1.12346
float4 | 1.12346
float8 | 1.12345678912346
```

Link to MySQL docs:
http://dev.mysql.com/doc/refman/5.5/en/floating-point-types.html

dimitri:
Thanks for the details! The precision is part of what's called the *typemod* in PostgreSQL slang, and it's available from the casting rules your patch is editing.

Please update your patch to target either real or float (single or double precision) in PostgreSQL depending on the precision in the MySQL float type used. See the `:typemod` clause as used in the the integer data types for inspiration.

dimitri:
Closing for clean-up. More work is needed (not much) to get accepted, per comments last year.

PR #406, opened 2016-05-13
add the postgres debian ppa key in the correct way

spaghetti-:
Travis builds are failing because the key can't be imported.


dimitri:
Thanks for having a look at the travis situation, I admit I couldn't set myself to act on that. Hope the experimental patch about using the debian provided SBCL version is going to help; I think it should as stable now has `1.2.4-2`.

spaghetti-:
I'll see if I can fix it up, got kinda caught up in stuff. I'm more used to CircleCI

PR #407, opened 2016-05-17
Add the “set not null” cast option for MySQL

KrzysiekJ:
Use case: Django dissuades setting `NULL` on text fields “on string-based fields […] because empty string values will always be stored as empty strings, not as `NULL`. If a string-based field has null=True, that means it has two possible values for »no data«: `NULL`, and the empty string. In most cases, it’s redundant to have two possible values for »no data«; the Django convention is to use the empty string, not `NULL`.”.

pgloader already supports custom transformations which can be used to replace `NULL` values in string-based columns with empty strings. Setting `NOT NULL` constraint on those columns could possibly be achieved by running a database query to extract their names and then generating relevant `ALTER TABLE` statements, but a cast option in pgloader is a more convenient way.

This pull requests lacks tests, as testing process in pgloader seems to be poorly documented and there is an encoding-related issue when running `make test` on my machine.

### DIFF
--- a/pgloader.1.md
+++ b/pgloader.1.md
@@ -1730,7 +1730,7 @@ The `database` command accepts the following clauses and options:
 	    The spelling *keep default* explicitly prevents that behaviour and
 	    can be used to overload the default casting rules.
 
-	  - *drop not null*, *keep not null*
+	  - *drop not null*, *keep not null*, *set not null*
 
         When the option *drop not null* is listed, pgloader drops any
         existing `NOT NULL` constraint associated with the given source
@@ -1739,6 +1739,10 @@ The `database` command accepts the following clauses and options:
 
 	    The spelling *keep not null* explicitly prevents that behaviour and
 	    can be used to overload the default casting rules.
+
+        When the option *set not null* is listed, pgloader sets a `NOT NULL`
+        constraint on the target column regardless whether it has been set
+        in the source MySQL column.
 
       - *drop typemod*, *keep typemod*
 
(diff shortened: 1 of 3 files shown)


dimitri:
Thanks a lot for your patch! For testing you need to have an UTF-8 compatible environment, which I force in spelling the command that way:

    LANG=EN_US.UTF-8 make check

Then again, the current testing we have does not cover database migration, only file based loading of a single target table at a time. It's in my mental TODO list to improve that, and I guess we could have Travis and others install and setup source MySQL database when needed, etc.

Finally, about the patch itself, you're covering all the source formats with CAST support here, that's SQLite, MS SQL and MySQL, in one go ;-)

Before I go merge the patch, can you confirm you did some local tests and it worked for you?

KrzysiekJ:
I have the `LANG` environment set to `en_US.UTF-8` (or `pl_PL.UTF-8`). I’ve reported the encoding issue in #408.

Changes in the code are adjacent to the fragments about `drop not null`. If `set not null` will work for all databases, does it mean that `drop not null` works that way too? It is put under the MySQL section in the documentation…

Local tests work for me (for the MySQL database :)).

dimitri:
Thanks for your patch, it's merged ;-)

You're right about `set not null` being available for all databases and the docs missing the memo... do you want to work on fixing that too?

KrzysiekJ:
I now see the “Please refer to the MySQL CAST clause for details.” information in other databases’ sections. (Previously I’ve been searching for fragments like “keep not null” and found them only in the MySQL section, so I assumed these options were not present in other databases). Are such references sufficient?

dimitri:
Yeah that should do. I'm still thinking some day I should split the huge man page into a main page and a page per source type, to make it easier for the users to read only what they need to and stop being confused. We can fix the _See also_ references when we split the page that way.

PR #410, opened 2016-05-18
Override encoding in every testing connection

### LINKED ISSUE #408: The “client_encoding” option doesn’t seem to be ignored
Running `make check` (after doing `make prepare` inside the `test/` folder) yields the following result:

```
2016-05-18T12:34:46.055000+02:00 WARNING pgloader always talk to PostgreSQL in utf-8, client_encoding has been forced to 'utf8'.
[…]
An unhandled error condition has been signalled:
   Database error 22P05: character with byte sequence 0xc4 0x99 in encoding "UTF8" has no equivalent in encoding "LATIN1"
QUERY: drop table if exists expected.allcols;

An unhandled error condition has been signalled:
   Database error 22P05: character with byte sequence 0xc4 0x99 in encoding "UTF8" has no equivalent in encoding "LATIN1"
QUERY: drop table if exists expected.allcols;
```

If I change the `SET client_encoding to 'latin1'` fragment in `allcols.load` into `SET client_encoding to 'utf8'`, then I get no encoding errors related to this particular file.

The description of commit a3fd22acd38430 states that a WARNING was added “whenever the client_encoding is set to a known bad value. It's a WARNING because we then simply force 'utf-8' anyway.” It seems that `utf-8` isn’t actually forced. Additionally, when I revert encoding value in a load file to `latin1`, the problem doesn’t come back — `make check` fails on a nearest untouched file with `latin1` encoding set. Apparently the testing environment isn’t purged properly between test runs (that may explain why this problem got unnoticed).

KrzysiekJ:
Fix #408.

### DIFF
--- a/src/main.lisp
+++ b/src/main.lisp
@@ -440,7 +440,8 @@
            (expected-data-file     (make-pathname :defaults load-file
                                                   :type "out"
                                                   :directory expected-subdir))
-           ((target-conn *pg-settings*) (parse-target-pg-db-uri load-file))
+           ((target-conn gucs) (parse-target-pg-db-uri load-file))
+           (*pg-settings* (pgloader.pgsql:sanitize-user-gucs gucs))
            (*pgsql-reserved-keywords* (list-reserved-keywords target-conn))
            (target-table (create-table (pgconn-table-name target-conn)))
 


dimitri:
Thanks for working on that, your fix looks like the right approach or abouts. I think it would be better to change the `*pg-settings*` binding from the `process-regression-test` entry point, something in the spirit of the non-tested written in the text area snippet here:

```diff
-           ((target-conn *pg-settings*) (parse-target-pg-db-uri load-file))
+           ((target-conn gucs) (parse-target-pg-db-uri load-file))
+           (*pg-settings* (pgloader.pgsql:sanitize-user-gucs gucs))
```

KrzysiekJ:
I’ve updated the pull request according to the snippet and also reduced the number of connections as suggested in #408 (had to find out what those surrounding asterisks mean in Common Lisp ;)). The `load-data` invocation (which loads expected data set) however has to be kept between creating expected data schema and comparing results, so I’m not convinced whether it is actually a better solution.

dimitri:
The `load-data` section should be left out of any `with-pgsql-connection` as it will connect at least once and in most cases several times on its own. What can be moved is the `drop` then `create` of the *expected* table in the final connection to compare the result, thus sparing a connection. I think it's worth it doing that.

KrzysiekJ:
But the `load-data` section has comment “load expected data”, so it needs to be executed after creating expected schema (after `drop` and `create`). Therefore the order of operations cannot be changed.

dimitri:
Damn you're right. I somehow missed that with the part that does the test itself. Sorry about that, I don't think we can easily simplify things here given how we re-use the internal APIs, and I think we should re-use internal APIs as much as possible. What we should do here is give the two SQL commands as a `:before` argument to the `load-data` bits.

KrzysiekJ:
If I understand correctly, the difficulty here derives from the fact that those SQL commands are generated inside the `with-schema` macro, which expects to be run inside a connection (that we’re trying to remove). I’ve updated the pull request, removing the code reorganization.

dimitri:
Thanks for the rewriting and sorry for the extra useless work!

KrzysiekJ:
No problem, it’s a pleasure to work with Lisp anyway. ;)

Issue #320, opened 2015-12-21
fails to separate similar tablenames from mysql


zesoup:
Howdy!

It seems to me that `mysql-schema.lisp:210` doesnt force case-sensitivity, regardless of quote identifiers. This applies to default settings of a current debian-jessie mysql. I'm not sure if forcing to a case-sensitive collation is the best idea though.

==MySQL==
```
Create Table A(id int);
Create Table a(id int);
```

==PGLoader==
```
QUERY: CREATE TABLE a
(
  id                     bigint,
  id                     bigint
);
```

dimitri:
Trying to reproduce here but my version of MySQL won't let me...

```
Create Table `a`(id int);
ERROR 1050 (42S01): Table 'a' already exists
```

dimitri:
Please retry with current code, I think it's been fixed with the refactoring of the schema introspection...

dimitri:
Should be fixed, closing. Please re-open if you still have the issue.

zesoup:
Reopening, as i still see the issue upstream. Pullrequest with enforced casesensitivity exists. I'm not entirely convinced that this whole casesensitivity issue is fixed, but it seems like this one is.

used options, as i havent provided this before:

load database
     from      mysql://name@host/database  
     into postgresql:///mysqlclone
 
 WITH create tables, no truncate, quote identifiers, reset no sequences;

dimitri:
Meanwhile I've been told that to reproduce this issue one need to use a case sensitive File System, so that the files backing up the tables `a` and `A` are not the same on disk... Does MySQL have a notion of an identifier for its table? I could easily use that to disambiguate here.

PR #419, opened 2016-06-08
MySQL schema-join now case insensitive. #320

zesoup:
In #320 it was reported that pgloader fails to enforce case insensitivity on MySQL imports.

Attached Patch seems to fix this, by forcing case sensitivity when joining tables with their definitions.
Only applies to case insensitive installations of mysql, which seem to be most of 'em.


dimitri:
The current understanding of #320 is about the file system case sensitivity. To reproduce you need to test on a case sensitive file system and witness that tables `a` and `A` are only then separate entities. I'm not sure your patch helps much what that scenario?

PR #473, opened 2016-11-06
Use uiop:run-program ... :directory ...

pipping:


dimitri:
Thanks! Can you tell me which version of uiop (or rather ASDF) is the first one to come with support for `:directory`? I might have merged too hastily, don't want to make it harder than necessary to build pgloader...

pipping:
**TL;DR**: I think you're fine.

I've just done some digging and ended up at [this commit](https://gitlab.common-lisp.net/asdf/asdf/commit/4caf0cc729b2db7e2e48155abf028ddb370c6864):

```
commit 4caf0cc729b2db7e2e48155abf028ddb370c6864
Author: Francois-Rene Rideau <tunes@google.com>
Date:   Sat Sep 28 14:00:32 2013 -0400

    Another reimplementation of run-program. New:
    * it handles INPUT and ERROR-OUTPUT, not just OUTPUT.
    * it accepts PATHNAMEs as I/O designators
    * it works on ECL under SLIME (fixes lp#1232225)
    * SLIGHT API CHANGE!
      At the suggestion of stassats, we're always returning an exit code,
      and in a predictable return value position.
      However, both for backward compatibility and usability,
      the OUTPUT slurping results remain the primary return value,
      followed by the ERROR-OUTPUT slurping results, followed by the exit code.
      So far, no one was using multiple values in SLURP-INPUT-STREAM methods,
      so no one will regret that, except people who might have somehow expected
      the old weird behavior of a return code when OUTPUT is NIL or :INTERACTIVE.
    * As another API tweak and improvement, keyword options are now specified
      separately for each of your I/O designators, by putting them your designator
      in a list. A list also distinguishes a string as input from a namestring
      to designate a pathname, i.e. :input '("bar") is the string literal,
      whereas :input "bar" is the file named bar.
    
    Several fixes and extensions to stream support.
    
    Only tested so far on Linux x64 with
    ccl clisp sbcl ecl ecl_bytecodes cmucl abcl allegro allegromodern xcl
```

from which on I would assume `run-program :directory` to work (because `%invoke-run-program` declared it accepted `:directory` and `run-program` always had `&allow-other-keys` in its signature).

`git describe --tags` for that revision gives me 3.0.2.9-9-g4caf0cc so 3.0.3 and anything after that should be fine. 3.0.3 is a very early release of asdf-3 so I think it's safe to say anything that has asdf-3 (which you need for uiop anyway) has asdf-3.0.3 or better. Moreover, this quote from the commit message:
```
      At the suggestion of stassats, we're always returning an exit code,
      and in a predictable return value position.
```
suggests that with asdf-3 prior to 3.0.3 the current `multiple-value-bind` probably would not have worked as intended either.

As a sidenote, I think the only way to end up without asdf-3 right now, I believe, is either
 - because you're on the free personal edition of lispworks (which is stuck at 6.x while all the paid versions are at 7.x and have asdf 3.x) or
 - because you're on the latest release of clisp, namely 2.49, (which is from 2011, I believe?). A hg checkout of clisp does contain asdf-3, the just don't have the manpower to push out a new release.

dimitri:
Thanks for digging the information, I see I shouldn't worry ;-)

PR #515, opened 2017-02-12
Shadow symbols NAMESTRING, NUMBER and INLINE in pgloader.parser package

scymtym:
Defining rules on standard symbols like ``cl:namestring`` is a bad idea since other systems may do the same, inadvertently overwriting each other's rules.

Furthermore, future esrap versions will probably prevent defining rules whose names are symbols in locked packages, making this change mandatory.


dimitri:
Thanks a lot @scymtym !

PR #572, opened 2017-06-09
MSSQL: Add support for typemods of various text types

disco-stu:
If a custom CAST rule is defined (e.g CAST type varchar to varchar) the
original typemod get's lost. This commit is based on Dimitris patch from
571 and adds typmod support for "char", "nchar", "varchar", "nvarchar"
and "binary".


dimitri:
Thanks for the testing and updated patch!

PR #573, opened 2017-06-09
MSSQL: fix typmod conversion for "max" typemods

disco-stu:
The previous commit makes it possible to convert typemods for various
text types. In MSSQL it's possible to create a column like varchar(max).
Internally this is reported as varchar(-1) which results in a CREATE
TABLE statement that contains e.g. varchar(-1).

This patch drops the typemod if it's -1 (max).

It's based on Dimitris patch slightly modified by myself.

### DIFF
--- a/src/sources/mssql/mssql-cast-rules.lisp
+++ b/src/sources/mssql/mssql-cast-rules.lisp
@@ -106,11 +106,16 @@
                           type
                           (mssql-column-numeric-precision col)
                           (or (mssql-column-numeric-scale col) 0)))))
-          ((member type '("char" "nchar" "varchar" "nvarchar" "binary") :test #'string=)
+
+          ((member type '("char" "nchar" "varchar" "nvarchar" "binary")
+                   :test #'string=)
            ;; the user might have a CAST rule with keep typemod, so we need
            ;; to deal with character-maximum-length for now
-           (format nil "~a(~a)"
-                   type (mssql-column-character-maximum-length col)))
+
+           (if (= -1 (mssql-column-character-maximum-length col))
+               type
+               (format nil "~a(~a)"
+                       type (mssql-column-character-maximum-length col))))
 
           (t type))))
 


dimitri:
Common Lisp has plenty of different equality operators. To make a long story short, for numbers we should use `=`. The `eq` operator tests something likes pointer equality and is mostly useful for known symbols and other things not *Garbage Collected*.

Can you update your patch using `=` rather than `eq` please?

disco-stu:
Ok, Thanks. I've updated the patch.

dimitri:
Thanks perfect! (and I assume you have tested it too)

disco-stu:
Yes, I've tested the patch. Here is what I've tested so far:

MSSQL:
```
CREATE TABLE [dbo].[tbl2](
	[id] [bigint] NULL,
	[c_varchar_50] [varchar](50) NULL,
	[c_nvarchar_50] [nvarchar](50) NULL,
	[c_text] [text] NULL,
	[c_ntext] [ntext] NULL,
	[c_char_50] [char](50) NULL,
	[c_nchar_50] [nchar](50) NULL,
	[c_varchar_max] [varchar](max) NULL,
	[c_nvarchar_max] [nvarchar](max) NULL
) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY]
```

CASTS:
```
CAST type varchar to varchar,
     type nvarchar to varchar,
     type char to char,
     type nchar to char
```

PGSQL:
```
CREATE TABLE tbl2 (
    id bigint,
    c_varchar_50 character varying(50) DEFAULT NULL::character varying,
    c_nvarchar_50 character varying(50) DEFAULT NULL::character varying,
    c_text text,
    c_ntext text,
    c_char_50 character(50) DEFAULT NULL::bpchar,
    c_nchar_50 character(50) DEFAULT NULL::bpchar,
    c_varchar_max character varying,
    c_nvarchar_max character varying
);
```

And PSQL without CASTS:
```
CREATE TABLE tbl2 (
    id bigint,
    c_varchar_50 text,
    c_nvarchar_50 text,
    c_text text,
    c_ntext text,
    c_char_50 text,
    c_nchar_50 text,
    c_varchar_max text,
    c_nvarchar_max text
);
```

This should cover the most new stuff.

dimitri:
Be careful that `bpchar` in PostgreSQL is actually black padded character type:
 https://www.postgresql.org/docs/current/static/datatype-character.html#DATATYPE-CHARACTER-SPECIAL-TABLE
 https://www.postgresql.org/message-id/6312.979871228@sss.pgh.pa.us

In general it's best to avoid it!

PR #574, opened 2017-06-10
Repair Travis CI builds

cbandy:
Had to uninstall the existing PostgreSQL instance and install a recent version of SBCL.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -30,4 +30,21 @@ postgresql_uninstall() {
 	packages
 }
 
+remote_file() {
+	local target="$1" origin="$2" sum="$3"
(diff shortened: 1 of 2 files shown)


dimitri:
Thank a lot for that work! Would you be up to testing either the default Travis PostgreSQL version or the current one, PostgreSQL 9.6, in our testing suite?

cbandy:
Absolutely. Do you want to test more than one version, or just the latest?

dimitri:
We could test both the Travis (Debian) default and the latest PostgreSQL stable version, which is 9.6 today. If that's also not too much worry I would like to test the current PostgreSQL beta version, currently PostgreSQL 10, which is available in the apt.pg.o repository of course.

PR #575, opened 2017-06-10
Multiple Travis CI jobs

cbandy:
Sets up one job for each combination of Lisp and PostgreSQL we want to test. I tried adding `PGVERSION=10` as discussed in https://github.com/dimitri/pgloader/pull/574#issuecomment-307563027, but there's no `postgresql-10-ip4r` package in the PGDG repositories, yet.

I added a commit that uses `ccl` even though the tests are failing. We can omit it if you like.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -2,6 +2,34 @@
 
 set -eu
 
+lisp_install() {
+	case "$LISP" in
+		ccl)
+			ccl_checksum='08e885e8c2bb6e4abd42b8e8e2b60f257c6929eb34b8ec87ca1ecf848fac6d70'
+			ccl_version='1.11'
+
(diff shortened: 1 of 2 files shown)


dimitri:
I see that the CCL build fails because of a memory error, much like in the docker hub environment.

```
Fatal error in "pgloader" : Fault during read of memory address #xD8E4
```

The odds are low that it will actually help, can we still setup `DYNSIZE` to like 512 in the CCL build case? it's an environment variable you pass to the make command (`make DYNSIZE=512 ...`).

Thanks for your work, I hope the CCL coverage will be possible!

cbandy:
No luck: https://travis-ci.org/cbandy/pgloader/builds/241651690

dimitri:
Thanks for trying, let's commit your patch with the whole matrix of test and have an eye from the CCL community have a look, then.

Issue #576, opened 2017-06-12
MSSQL: Wrong translation of default value 'getdate()'


disco-stu:
The current implementation translates the default value getdate() to 'now' which leads to CREATE TABLE statements like:

```
CREATE TABLE dbo.tbl_default_test 
(
  id      bigint,
  created timestamptz default 'now'
);
```
Because PostgreSQL evaluates 'now'  the created table uses the current timestamp:

```
\d db.otbl_default_test
                                    Table "dbo.tbl_default_test"
 Column  |           Type           |                             Modifiers                             
---------+--------------------------+-------------------------------------------------------------------
 id      | bigint                   | 
 created | timestamp with time zone | default '2017-06-12 18:09:36.449082+02'::timestamp with time zone
```

I'll come up with a possible fix.

disco-stu:
#577 contains a possible solution.

dimitri:
Thanks for the solution, another option would have been using default 'now()'. Not that your patch works thanks to https://github.com/dimitri/pgloader/blob/master/src/sources/mssql/mssql-cast-rules.lisp#L149 ; so we could also have added 'now' as a special default value there.

disco-stu:
Yes, I also considered this option. But it seems that all other sources are using CURRENT_TIMESTAMP. However, I've tried to keep the change as small as possible. Moreover the current solution doesn't touch anything outside MSSQL.
Thanks for merging.

PR #577, opened 2017-06-12
MSSQL: Fix Default value translation for getdate() (#576)

disco-stu:
Currently the default value getdate() is replaced by 'now' which creates
statements like:
  CREATE TABLE ... created timestamp 'now' ...
which leads to table definitions like:
  default '2017-06-12 17:54:04.890129'::timestamp without time zone

This is because 'now' is evaluated when creating the table.

This commit fixes the issue by using CURRENT_TIMESTAMP as default
instead of 'now'.


dimitri:
Thanks!

PR #578, opened 2017-06-12
Debian

disco-stu:
Lintian cries about the compat level as well as the standards-version.


dimitri:
Which Debian version did you build the package in? Need to refresh my memories here but I think we need to be careful and place in the values for the current Sid system (of course @ChristophBerg would know for sure). I usually see about those changes at release time, but I guess it's good to maintain them anyway.

disco-stu:
I'm using debian stretch but build within a sid chroot. I've check the debhelper changelog and didn't find any reason to not bump compat and standsards-version.

Compat level 8 is deprecated. One could think about bumping it to 10 but I didn't check possible breakages.

@ChristophBerg any thoughts?

dimitri:
Release time is approaching fast, thanks for the debian/ update!

PR #580, opened 2017-06-12
MSSQL: Add ON UPDATE / DELETE support for fkeys

disco-stu:
The former query to find foreign key constraints doesn't consider ON
UPDATE and ON DELETE rules.


dimitri:
Thanks for the patch! 

Merged assuming you tested it locally of course, and seeing that the travis tests are passing for SBCL -- we still have an environment / architecture problem with Travis for the CCL tests.

PR #581, opened 2017-06-15
Fix the brew install command in README

pmatseykanets:
If you run 

```
brew install --head pgloader
```

the brew will error out with the message

```
Error: Specify `--HEAD` in uppercase to build from trunk.
```


dimitri:
Thanks!

PR #606, opened 2017-08-07
Ignore nulls in varbinary-to-string transform

lsnape:
I've tested this on one of my import jobs and can confirm that the rows containing a `NULL` `binary(n)` value are preserved.


dimitri:
Thanks!

Issue #651, opened 2017-10-20
Don't work if MS SQL database in case sensitive collation.


Splarv:
```
$ dpkg -s pgloader 
Version: 3.4.1+dfsg-1~pgdg17.04+1
```
MS SQL 2008
Collation Cyrillic_General_CS_AI

Log

```
sb-impl::*default-external-format* :UTF-8
tmpdir: #P"/tmp/pgloader/"
2017-10-20T15:30:53.014000Z NOTICE Starting pgloader, log system is ready.
2017-10-20T15:30:53.026000Z INFO Starting monitor
2017-10-20T15:30:53.234000Z DEBUG CONNECTED TO #<PGLOADER.PGSQL:PGSQL-CONNECTION pgsql://olleg@UNIX:5432/pakdb {10098B2953}>
2017-10-20T15:30:53.234000Z DEBUG SET client_encoding TO 'utf8'
2017-10-20T15:30:53.234000Z DEBUG SET application_name TO 'pgloader'
2017-10-20T15:30:53.234000Z SQL MSSQL: sending query: -- params: dbname
--         table-type-name
--         including
--         filter-list-to-where-clause including
--         excluding
--         filter-list-to-where-clause excluding
  select c.table_schema,
         c.table_name,
         c.column_name,
         c.data_type,
         CASE
         WHEN c.column_default LIKE '((%' AND c.column_default LIKE '%))' THEN
             CASE
                 WHEN SUBSTRING(c.column_default,3,len(c.column_default)-4) = 'newid()' THEN 'generate_uuid_v4()'
                 WHEN SUBSTRING(c.column_default,3,len(c.column_default)-4) LIKE 'convert(%varchar%,getdate(),%)' THEN 'today'
                 WHEN SUBSTRING(c.column_default,3,len(c.column_default)-4) = 'getdate()' THEN 'CURRENT_TIMESTAMP'
                 WHEN SUBSTRING(c.column_default,3,len(c.column_default)-4) LIKE '''%''' THEN SUBSTRING(c.column_default,4,len(c.column_default)-6)
                 ELSE SUBSTRING(c.column_default,3,len(c.column_default)-4)
             END
         WHEN c.column_default LIKE '(%' AND c.column_default LIKE '%)' THEN
             CASE
                 WHEN SUBSTRING(c.column_default,2,len(c.column_default)-2) = 'newid()' THEN 'generate_uuid_v4()'
[...]

PR #653, opened 2017-10-24
Fix issue #651

Splarv:
https://github.com/dimitri/pgloader/issues/651


dimitri:
I would like to squash/merge the commits separately so that I can rewrite the commit message to my liking, but it seems I can't do that with both the commits in the same PR. Can you split them?

dimitri:
I did it the simple way then, thanks!

PR #689, opened 2017-12-01
Hack to allow unique indexes to convert to db with btree_gin installed

bmdavi3:
Almost certainly not a proper fix, but this is what we needed to do recently when migrating a MySQL database to a PostgreSQL instance with the btree_gin extension installed.

To reproduce...

MySQL:

`CREATE TABLE foo (bar varchar(20), constraint bar_uniq unique (bar));`

PostgreSQL:

`create extension btree_gin;`

Converting with pgloader produces the following error:

```
2017-12-01T00:31:49.841000Z ERROR PostgreSQL Database error 0A000: access method "gin" does not support unique indexes
QUERY: CREATE UNIQUE INDEX idx_1108501_bar_uniq ON test_db.foo USING gin(bar);
```

Thanks for all the work on pgloader, hopefully the proper fix is obvious for someone more familiar with the system catalogs.

### DIFF
--- a/src/pgsql/sql/list-typenames-without-btree-support.sql
+++ b/src/pgsql/sql/list-typenames-without-btree-support.sql
@@ -12,4 +12,5 @@ select typname,
                 join pg_type t on c.opcintype = t.oid
           where amname = 'btree' and t.oid = pg_type.oid
        )
+       and typname not in ('varchar', 'cidr')
 group by typname;


dimitri:
After some investigation as reported below, what happens is that the `varchar` datatype doesn't have a btree operator class defined proper. Only `text` has those definitions. What happens if you cast the `varchar(20)` to `text` in your case?

I am not a fan of blacklist `varchar` (and `cidr`) from the type names without btree support list we fetch at runtime.

Here's the query I've been using to investigate on the situation, after installing `btree_gin` too:

```sql
select amname, c.opcname, t.typname
  from pg_am am
       join pg_opclass c on am.oid = c.opcmethod
       join pg_type t on c.opcintype = t.oid
 where t.oid = 'text'::regtype
    or t.oid = 'varchar'::regtype;
```

Which gives: 

```
 amname |       opcname       | typname 
--------+---------------------+---------
 btree  | text_ops            | text
 hash   | text_ops            | text
 btree  | varchar_ops         | text
 hash   | varchar_ops         | text
 btree  | text_pattern_ops    | text
 btree  | varchar_pattern_ops | text
 hash   | text_pattern_ops    | text
 hash   | varchar_pattern_ops | text
 spgist | text_ops            | text
 brin   | text_minmax_ops     | text
 gist   | gist_text_ops       | text
 gin    | text_ops            | text
 gin    | varchar_ops         | varchar
(13 rows)
```

bmdavi3:
That does indeed fix the error.  I'm casting like this...

```
LOAD DATABASE
     FROM      mysql:///test_db
     INTO postgresql:///test_db

CAST type varchar to text drop typemod;
```

And it creates the unique index just fine.

PR #711, opened 2018-01-09
doc: fix a few link format

jd:
They are still in Markdown format, remove or move to rst.


dimitri:
Thanks! The travis failure is on me, I failed to realise the problem before, I broke it with a previous patch (https://github.com/dimitri/pgloader/commit/81be9ae60ed63f3b2c10c35d159215dc857b0ace). I'll have a look soon.

PR #761, opened 2018-03-12
Fix minor error in docs about logfile location

chosak:
The default logfile location seems to be `/tmp/pgloader/pgloader.log`, not `/tmp/pgloader.log` as currently documented. This is observable in practice and also in [the source code](https://github.com/dimitri/pgloader/blob/5b227200a9525050ce60e44e2bc899a3c2b1f018/src/main.lisp#L110).


dimitri:
Thanks!

PR #771, opened 2018-03-26
add hex-to-dec

kxjhlele:
add hex-to-dec

### DIFF
--- a/src/utils/transforms.lisp
+++ b/src/utils/transforms.lisp
@@ -374,5 +374,7 @@
     (null    nil)
     (string (base64:base64-string-to-string string))))
 
-(defun hex-to-dec (hex)
-  ( write-to-string (parse-integer hex :radix 16)))
+(defun hex-to-dec (string)
+  (typecase hex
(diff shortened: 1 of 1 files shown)


dimitri:
Nice addition, thanks!

Before merging the PR though, please review some of the other functions that do that kind of things, such as `base64-decode` where we check the input type and nicely handle NIL values (if NULL) type, and also thanks to `etypcase` error out early rather than trying to parse a non-string value.

I expect you will end up with something that looks more like `integer-to-string`.

The reason of the complexity of having to deal with either strings or integers is that it depends on the source driver what you get here...

dimitri:
Thanks for the update. A last improvement to make: rather than `typecase` can you please use `etypecase` here? The former will return null if the input type is neither NULL nor string, meaning we're silently loosing data. The latter will signal a condition that will be logged and at least users can react to that.

There's also the documentation to edit, but I can take care of that if you want.

dimitri:
Thanks!

PR #815, opened 2018-07-20
>= 10 instead of <= 10

alexknips:
the given explanation seemed wrong


dimitri:
Thanks!

alexknips:
Do you also need to fix it in code or is the code correct?

dimitri:
The code is correct, have a look at https://github.com/dimitri/pgloader/blob/master/src/sources/mysql/mysql-cast-rules.lisp#L25 and around.

alexknips:
got it. not that familiar with LISP. Thx for the clarification.

Issue #588, opened 2017-06-27
Problems installing on Mac OSX


lennartkramer1988:
Hey,

I'm trying to install pgloader on Mac OSX El Capitan but brew is giving me problems. pgloader has a dependency to postgresql8 and this formula seems to have been removed from brew. 

Please advice on how to get pgloader running properly.

Regards,

Lennart

```
~ brew install pgloader
Error: No available formula with the name "postgresql8" (dependency of pgloader)
==> Searching for a previously deleted formula...
Error: No previously deleted formula found.
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.

~ brew install postgresql8
Error: No available formula with the name "postgresql8"
==> Searching for a previously deleted formula...
Error: No previously deleted formula found.
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```

dimitri:
Maybe you need to upgrade brew or something, I don't see such a dependency here. Also, I am given the option to install without postgresql (which could be remote after all).

```
~ brew info pgloader
pgloader: stable 3.3.2 (bottled), HEAD
Data loading tool for PostgreSQL
https://github.com/dimitri/pgloader
Not installed
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/pgloader.rb
==> Dependencies
Build: buildapp ✘
Required: sbcl ✔, freetds ✘
==> Requirements
Recommended: postgresql ✔
==> Options
--without-postgresql
	Build without postgresql support
--HEAD
	Install HEAD version
```

lennartkramer1988:
Hey Dimitri,

Installing without dependencies worked, thanks.

Command below fixed my problem:
```
brew install pgloader --ignore-dependencies coq
```
Regards,

Lennart

shahrukh-imran:
c

PR #819, opened 2018-08-03
Added how to install w/brew w/o postgresql server dependency

jungblud:
Handy standalone install using Homebrew without postgresql server dependency. Handy if using another approach to where your PostgreSQL server lives (like https://github.com/PostgresApp/PostgresApp) and you want to keep things tidy! This was an issue that led to this solution: https://github.com/dimitri/pgloader/issues/588

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ it builds aginst current set of dependencies versions.
 ### Building from sources on macOS
 
 When using [brew](https://brew.sh), it should be a simple `brew install
---HEAD pgloader`.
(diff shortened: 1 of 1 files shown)


tbussmann:
> using Homebrew without postgresql server dependency

Why do you recommend to ignore _all_ dependencies in that case? Isn't this what pgloader's brew formula option `--without-postgresql` explicitly is designed for? Also, why do you include the installation of `coq` for this use case?

df7cb:
Closing as by last comment.

PR #821, opened 2018-08-09
Improved Dockerfiles/docker image size

uniquestring:
Reduced size of the docker images by using smaller base image and multistage builds:

```
dimitri/pgloader   latest       452MB
pgloader           latest       171MB

dimitri/pgloader   ccl.latest   730MB
pgloader           ccl.latest   248MB

```

Also reduced layer count of the images.

We could save even more space by installing as few packages as possible. You don't need `make` etc. in the final container. Installing less packages inside the builder container could also improve build time.

I think the package requirements in README are outdated. Some packages are missing and some others are only required during compilation.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,5 @@
 .git
 .vagrant
 build
+Dockerfile
+Dockerfile.ccl
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,30 @@
 FROM debian:stretch
-MAINTAINER Dimitri Fontaine <dim@tapoueh.org>
 
-RUN apt-get update                                   && \
-    apt-get install -y --no-install-recommends          \
-                    wget curl make git bzip2 time       \
-                    ca-certificates                     \
-                    libzip-dev libssl1.1 openssl        \
-                    patch unzip libsqlite3-dev gawk     \
-                    freetds-dev sbcl                 && \
-    rm -rf /var/lib/apt/lists/*
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+      bzip2 \
+      ca-certificates \
+      curl \
+      freetds-dev \
+      gawk \
+      git \
+      libsqlite3-dev \
+      libssl1.1 \
+      libzip-dev \
+      make \
+      openssl \
+      patch \
+      sbcl \
+      time \
+      unzip \
+      wget \
+    && rm -rf /var/lib/apt/lists/*
 
-ADD ./ /opt/src/pgloader
-WORKDIR /opt/src/pgloader
+COPY ./ /opt/src/pgloader
 
-# build/ is in the .dockerignore file, but we actually need it now
-RUN mkdir -p build/bin
-RUN make
+RUN mkdir -p /opt/src/pgloader/build/bin \
+    && cd /opt/src/pgloader \
+    && make \
+    && mv /opt/src/pgloader/build/bin/pgloader /usr/local/bin
 
-RUN cp /opt/src/pgloader/build/bin/pgloader /usr/local/bin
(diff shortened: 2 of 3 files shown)


dimitri:
Thanks a lot for your contribution! If you want to work on reducing the space even further, I would appreciate that!

uniquestring:
I would love to do so, but I think I'll need your help. Could you add an list of dependencies, that are needed to run this application? Without installing the dependencies from the README, the program doesn't work. However I don't think that we need all of them to **run** this application (`make`  for example).

I could just guess or install packages one by one until the application _seems_ to work, but I want to avoid errors that are not immediately apparent or only become visible when running the application in a specific scenario.

The Ubuntu repo lists `libc6 zlib1g freetds-dev` as dependencies for their pgloader-package but - as usual - it's not up to date (3.4.1). Installing just those packages + the pgloader binary doesn't work:

<details>
  <summary>Output when running pgloader</summary>

```
debugger invoked on a SIMPLE-ERROR in thread
#<THREAD "main thread" RUNNING {1005D363F3}>:
  Error opening shared object "libsqlite3.so.0":
  libsqlite3.so.0: cannot open shared object file: No such file or directory.

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [CONTINUE       ] Skip this shared object and continue.
  1: [RETRY          ] Retry loading this shared object.
  2: [CHANGE-PATHNAME] Specify a different pathname to load the shared object from.

(SB-SYS:DLOPEN-OR-LOSE #S(SB-ALIEN::SHARED-OBJECT :PATHNAME #P"libsqlite3.so.0" :NAMESTRING "libsqlite3.so.0" :HANDLE NIL :DONT-SAVE NIL))
0] ^C

debugger invoked on a SB-SYS:INTERACTIVE-INTERRUPT in thread
#<THREAD "main thread" RUNNING {1005D363F3}>:
  Interactive interrupt at #x7F1C163B3660.

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [CONTINUE       ] Return from SB-UNIX:SIGINT.
  1: [ABORT          ] Reduce debugger level (to debug level 1).
  2:                   Skip this shared object and continue.
  3: [RETRY          ] Retry loading this shared object.
  4: [CHANGE-PATHNAME] Specify a different pathname to load the shared object from.

("bogus stack frame")
0[2] ^C

debugger invoked on a SB-SYS:INTERACTIVE-INTERRUPT in thread
#<THREAD "main thread" RUNNING {1005D363F3}>:
  Interactive interrupt at #x7F1C163B3660.

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [CONTINUE       ] Return from SB-UNIX:SIGINT.
  1: [ABORT          ] Reduce debugger level (to debug level 2).
  2:                   Return from SB-UNIX:SIGINT.
  3:                   Reduce debugger level (to debug level 1).
  4:                   Skip this shared object and continue.
  5: [RETRY          ] Retry loading this shared object.
  6: [CHANGE-PATHNAME] Specify a different pathname to load the shared object from.

("bogus stack frame")
0[3] ^C
```
This goes on everytime you press ctrl + c. YOu need to manually kill the container.
</details>

dimitri:
Yeah C-c in the debugger isn't very useful, is it? Here's the docs for using it: http://www.sbcl.org/manual/#Exiting-Commands-1; so you might just type in `continue` or exit the debugger with C-d, as when you want to exit a shell.

Other than that, SQLite is loaded by pgloader, and openssl too. The debian packaging in this repository is actually maintained and you can read https://github.com/dimitri/pgloader/blob/master/debian/control for the details.

Issue #844, opened 2018-10-04
pgloader freezes when it encounters a real type in MSSQL


ftzdomino:
- [ ] pgloader --version
  
    ```
source d356bd501b557b41502a45ac7471e055112962f5
    ```
  
  - [x] did you test a fresh compile from the source tree?
    
Yes
    
  - [ ] did you search for other similar issues?
Yes
    
  - [ ] how can I reproduce the bug?
Create any MSSQL database with a "real" data type.

The following shell script (docker required) will recreate the issue:
```
#!/bin/sh
set -e
MSSQL_PASSWORD=`echo | mkpasswd -s`
echo "Generated password: ${MSSQL_PASSWORD}"
DOCKER_INSTANCE_NAME="pgloader_test_$$"
docker run -e 'ACCEPT_EULA=Y' -e SA_PASSWORD=${MSSQL_PASSWORD} -p 0:1433 --name ${DOCKER_INSTANCE_NAME} -d microsoft/mssql-server-linux:2017-latest
sleep 1

trap cleanup 1 2 3 6

cleanup()
{
    docker stop ${DOCKER_INSTANCE_NAME}
    docker rm ${DOCKER_INSTANCE_NAME}
}


MSSQL_PORT=`docker inspect -f '{{ (index (index .NetworkSettings.Ports "1433/tcp") 0).HostPort }}' ${DOCKER_INSTANCE_NAME}`


TEST_SQL="CREATE DATABASE pgloader_test;"
docker exec -it ${DOCKER_INSTANCE_NAME} /opt/mssql-tools/bin/sqlcmd  -S localhost -U SA -P "${MSSQL_PASSWORD}" -s';' -h-1 -W -Q "${TEST_SQL}"


TEST_SQL="create table real_test (foo real);  insert into real_test (foo) values (1.33333333);"
docker exec -it ${DOCKER_INSTANCE_NAME} /opt/mssql-tools/bin/sqlcmd -S localhost -U SA -P "${MSSQL_PASSWORD}" -d "pgloader_test" -s';' -h-1 -W -Q "${TEST_SQL}"
PGLOADER_CONFIG_FILE=`mktemp`

echo "LOAD DATABASE
FROM mssql://SA:${MSSQL_PASSWORD}@localhost/pgloader_test
INTO postgresql://pgloader:pgloader@localhost/pgloader_test
with create tables, create schemas
;" > $PGLOADER_CONFIG_FILE
[...]

PR #845, opened 2018-10-04
Fix pgloader bug #844 by adding support for mssql real types

ftzdomino:
syb-real is a 4 byte float, which seems to correspond exactly to the lisp type float


dimitri:
Thanks a lot!

Issue #832, opened 2018-09-13
pgloader 3.5.2 fails to compile from source: Undefined instruction: MOVZXD


fxcoudert:
Compiling pgloader 3.5.2 from source, on macOS, with sbcl 1.4.11 leads to the following error:

```
==> make pgloader-standalone BUILDAPP=buildapp
buildapp    --require sb-posix --require sb-bsd-sockets --require sb-rotate-byte                        \
                       --sbcl sbcl                            \
                       --load-system pgloader               \
                       --load src/hooks.lisp                   \
                       --entry pgloader:main                   \
                       --dynamic-space-size 4096         \
                       --compress-core                    \
                       --output build/bin/pgloader
;; loading system "pgloader"
Fatal SIMPLE-ERROR:
  Compilation failed: Undefined instruction: MOVZXD in
 (INST MOVZXD RESULT MEMREF) in /private/tmp/pgloader-20180913-76774-1hh87zh/pgloader-3.5.2/lib/nibbles/sbcl-opt/x86-64-vm.lisp
make: *** [pgloader-standalone] Error 1
```

This issue was seen on all macOS versions in Homebrew (https://github.com/Homebrew/homebrew-core/issues/31591), and also reported to Debian so it also occurs on Linux: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=908641

svillemot:
I reported this issue on SBCL bug tracker: https://bugs.launchpad.net/sbcl/+bug/1792680

attila-lendvai:
it's a bug in the project called nibbles. it was using SBCL's internal API, and unsurprisingly it broke at a random point in time. it has been fixed in nibbles already:

https://github.com/sharplispers/nibbles/commit/8111b881179fe6d4c8fb9b314c2055cc15d78260

svillemot:
@attila-lendvai Thanks for the info.

fxcoudert:
@martinstreicher until a new release is cut, with the fixes inside, it won't be in Homebrew

martinstreicher:
AOK. I will just not use `—HEAD` and get an older version.

fxcoudert:
Can a new release please be made, containing the fix for sbcl 1.4.11 and later?

mguentner:
Running into the same issue. I am currently writing a derivation for `pgloader` in order to add it to NixOS.
Due to the way dependencies are handled by `pgloader`'s `Makefile`, I need to use the bundled version of it.
[...]

PR #853, opened 2018-10-10
Add Docker build instructions

lgebhardt:
Adds instructions for building Docker images using SBCL or CCL.

I encountered issues building the standard SBCL Dockerfile, possibly related to #832. It wasn't readily apparent that there was a ccl option for the Docker build. This should draw users' attention to this option.


dimitri:
Thanks for the patch @lgebhardt !

PR #854, opened 2018-10-12
add cl-ironclad and cl-babel dependencies to docker builds

jasonrig:
I had some issues building using the supplied Dockerfiles (I tried the CCL version only, in this case) that were resolved by adding two packages from apt: cl-ironclad and cl-babel. Contributing back the changes here.


dimitri:
Thanks for the fix @jasonrig !

PR #902, opened 2019-02-04
Allow "preserve index names" for MSSQL

ChristophKaser:
With this change, preserve index names is also supported for MSSQL-connections.

### DIFF
--- a/src/parsers/command-mssql.lisp
+++ b/src/parsers/command-mssql.lisp
@@ -31,6 +31,7 @@
                           option-create-tables
                           option-create-schemas
                           option-create-indexes
+			  option-index-names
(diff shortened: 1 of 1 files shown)


dimitri:
I'm surprised it's all that's needed, but a quick reading of the sources hints that it could well be. Just to make sure, how did you test the patch?

dimitri:
A lot of the support for those options is generic and written in the generic parts of the code. The parsing of the options is database-source specific though, and sometimes the generic part of the feature support is all that is needed. Long story short, thanks for the patch!

ChristophKaser:
Thank you for merging, and thank you for pgloader!

PR #907, opened 2019-02-13
Correctly parse PostgreSQL version string on Windows.

anhi:
The regex used in parse-postgresql-version-string failed to handle
"PostgreSQL 10.6, compiled by Visual C++ build 1800, 64-bit"
correctly.


dimitri:
Hi @anhi ; thanks for the patch, that's awesome!

We're only missing one piece here, and I would add it after pushing your patch if you want me to, but I though you might be happy to finish it yourself. Your patch is leaving a warning:

```
 PARSE-POSTGRESQL-VERSION-STRING
;Compiler warnings for "/Users/dim/dev/pgloader/src/pgsql/connection.lisp" :
;   In PARSE-POSTGRESQL-VERSION-STRING: Unused lexical variable MAYBE-OS
```

That's because we don't actually use the new variable `maybe-os`. We need to tell the compiler that we're not going to use it, and that's done in Common Lisp with a `declare`. The code would look like this with the proper declaration:

``` lisp
(cl-ppcre:register-groups-bind (full-version maybe-os maybe-variant)
      ("PostgreSQL ([0-9.]+)( [^,]+)?, [^,]+, (.*)" version-string)
    (declare (ignore maybe-os))
    (let* ((version-dots  (split-sequence:split-sequence #\. full-version))
...
```

Do you want to update your PR so that I can then merge?

anhi:
Hi @dimitri - thanks for the tip! I just pushed the corrected commit.

PR #956, opened 2019-04-30
Fixes for make rpm

stevenlevin:
Changes needed for make rpm on centos 6 / amazon linux

### DIFF
--- a/bootstrap-centos.sh
+++ b/bootstrap-centos.sh
@@ -7,9 +7,9 @@ sudo yum -y install yum-utils rpmdevtools @"Development Tools" \
 sudo yum -y install epel-release
 sudo yum install -y sbcl.x86_64 --enablerepo=epel
 
-wget http://downloads.sourceforge.net/project/sbcl/sbcl/1.3.6/sbcl-1.3.6-source.tar.bz2
-tar xfj sbcl-1.3.6-source.tar.bz2
-cd sbcl-1.3.6
+wget http://downloads.sourceforge.net/project/sbcl/sbcl/1.3.21/sbcl-1.3.21-source.tar.bz2
+tar xfj sbcl-1.3.21-source.tar.bz2
+cd sbcl-1.3.21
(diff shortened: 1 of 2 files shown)


dimitri:
If we're going to update the SBCL version we download to build the package, can we pick the current release, which now is 1.5.2 at http://www.sbcl.org ?

stevenlevin:
Sure.  I just did not test with that build.  I can test it later and resubmit.

stevenlevin:
RPM build failed on latest source with SBCL 1.5.2
`; caught ERROR:
;   READ error during COMPILE-FILE:
;
;     Symbol "CLOSE-MEMO" not found in the DB3 package.
;
;       Line: 26, Column: 17, File-Position: 816
;
;       Stream: #<SB-INT:FORM-TRACKING-STREAM for "file /root/rpmbuild/BUILD/pgloader/src/sources/db3/db3-schema.lisp" {10034B2A13}>`

tigerfoot:
Note that sbcl is now at 2.0.1

df7cb:
This has been resolved with other commits.

tigerfoot:
> For the sbcl version, I'd rather like to see something non-volatile that doesn't need updating every few weeks.

Sorry that's the pace sbcl upstream has decide and publish, openSUSE Tumbleweed being a rolling release we have to give developper the fresh meat.
Maybe I will just stop to fight to keep pgloader building for that platform :-/

df7cb:
Can't openSUSE just include an sbcl package that we can use?

tigerfoot:
What do you mean by sbcl package you can use, sbcl is available for Leap and Tumbleweed ? Sorry it seems I don't catch the issue.

df7cb:
If there's an OS-provided sbcl, we don't need to download it in https://github.com/dimitri/pgloader/blob/master/bootstrap-centos.sh#L12-L16

tigerfoot:
Yes Leap 15.3 and 15.4 has sbcl 2.1.x, Tumbleweed the rolling has always the last available actually 2.2.5 directly from the main repository 
zypper install sbcl 
do the trick.

PR #960, opened 2019-05-07
Remove invalid WITH default option

welcoMattic:
It seems like `create schema` option for WITH statement is invalid.


dimitri:
Thanks!

PR #968, opened 2019-05-16
Fix ccl docker images not having libssl

edcarter:
pgloader:ccl.latest throws an error: Shared library not open: "libssl.so.1.1". This commit adds libssl1.1 to the docker image which fixes the issue.


dimitri:
Thanks!

PR #970, opened 2019-05-18
Generic Function API for Materialized Views support.

dimitri:
Generalize support code for Materialized Views support, as it is about the same for the three databases where we have support for it.


src/sources/db3/db3-schema.lisp
```diff
@@ -4,40 +4,24 @@
 
 (in-package :pgloader.source.db3)
 
-(defclass dbf-connection (fd-connection)
-  ((db3 :initarg db3 :accessor fd-db3))
-  (:documentation "pgloader connection parameters for DBF files."))
+(defclass copy-db3 (db-copy)
+  ((encoding    :accessor encoding	  ; file encoding
+	        :initarg :encoding))
```
phoe: This line mixes tabs and spaces.

src/sources/mssql/mssql-connection.lisp
```diff
@@ -0,0 +1,48 @@
...
+(defmethod initialize-instance :after ((msconn mssql-connection) &key)
+  "Assign the type slot to mssql."
+  (setf (slot-value msconn 'type) "mssql"))
+
+(defmethod open-connection ((msconn mssql-connection) &key)
+  (setf (conn-handle msconn) (mssql:connect (db-name msconn)
+                                            (db-user msconn)
+                                            (db-pass msconn)
+                                            (db-host msconn)))
+  ;; apply mysql-settings, if any
```
phoe: Typo: you likely meant `mssql` here.

src/sources/mssql/mssql-schema.lisp
```diff
@@ -61,46 +39,55 @@
...
-                          (filter-list-to-where-clause including
-                                                       nil
-                                                       "c.table_schema"
-                                                       "c.table_name")
-                          excluding     ; do we print the clause?
-                          (filter-list-to-where-clause excluding
-                                                       t
-                                                       "c.table_schema"
-                                                       "c.table_name")))
+     (let ((including
```
phoe: `LET` inside `LOOP` is not idiomatic - local variables should be declared via the `:WITH` clause if they do not depend on the results of the iteration.

My suggestion that should be equivalent (but let's test it anyway):

```lisp
(defmethod fetch-columns ((catalog catalog)
                          (mssql copy-mssql)
                          &key
                            (table-type :table)
                            including
                            excluding
                          &aux
                            (table-type-name
                             (cdr (assoc table-type *table-type*))))
  (loop
    :with including := (filter-list-to-where-clause mssql
                                                    including
                                                    :not nil
                                                    :schema-col "c.table_schema"
                                                    :table-col "c.table_name")
    :with excluding := (filter-list-to-where-clause mssql
                                                    excluding
                                                    :not t
                                                    :schema-col "c.table_schema"
                                                    :table-col "c.table_name")
    :for (schema-name table-name name type default nullable identity
                      character-maximum-length
                      numeric-precision numeric-precision-radix numeric-scale
                      datetime-precision
                      character-set-name collation-name)
      :in (mssql-query (format nil
                               (sql "/mssql/list-all-columns.sql")
                               (db-name *mssql-db*)
                               table-type-name
                               including   ; do we print the clause?
                               including
                               excluding   ; do we print the clause?
                               excluding))
    :do
       (let* ((schema     (maybe-add-schema catalog schema-name))
              (table      (maybe-add-table schema table-name))
              (field
                (make-mssql-column
                 schema-name table-name name type default nullable
                 (eq 1 identity)
                 character-maximum-length
                 numeric-precision numeric-precision-radix numeric-scale
                 datetime-precision
                 character-set-name collation-name)))
         (add-field table field))
    :finally (return catalog)))
```

src/sources/mssql/mssql-schema.lisp
```diff
@@ -115,24 +102,30 @@
...
-                                                            "schema_name(schema_id)"
-                                                            "o.name"
-                                                            )
-                               excluding ; do we print the clause?
-                               (filter-list-to-where-clause excluding
-                                                            t
-                                                            "schema_name(schema_id)"
-                                                            "o.name"
-                                                            )))
+     :in  (let ((including
```
phoe: Ditto.

src/sources/mssql/mssql-schema.lisp
```diff
@@ -154,25 +147,32 @@
...
-                               (filter-list-to-where-clause including
-                                                            nil
-                                                            "kcu1.table_schema"
-                                                            "kcu1.table_name")
-                               excluding ; do we print the clause?
-                               (filter-list-to-where-clause excluding
-                                                            t
-                                                            "kcu1.table_schema"
-                                                            "kcu1.table_name")))
+     :in  (let ((including
```
phoe: Ditto.

src/sources/mysql/mysql-schema.lisp
```diff
@@ -4,57 +4,29 @@
...
-      (when views
-        (let ((sql
-               (format nil "DROP VIEW ~{~a~^, ~};"
-                       (mapcar (lambda (qname)
-                                 (format nil "~@[~s.~]~a"
-                                         (car qname) (cdr qname)))
-                               (mapcar #'car views)))))
-          (log-message :info "MySQL: ~a" sql)
-          (mysql-query sql))))))
+(defclass copy-mysql (db-copy)
```
phoe: Minor issue: if the superclass is named `db-copy` then the children should perhaps be named `mysql-copy`, `mssql-copy`, ...

`copy-mysql` sounds much more like a verb (function) than a noun (instance).

src/sources/mysql/mysql-schema.lisp
```diff
@@ -4,57 +4,29 @@
...
+             :initform nil)
+   (range-list :accessor range-list
+               :initarg :range-list
+               :initform nil))
+  (:documentation "pgloader MySQL Data Source"))
+
+(defmethod initialize-instance :after ((source copy-mysql) &key)
+  "Add a default value for transforms in case it's not been provided."
+  (let ((transforms (and (slot-boundp source 'transforms)
+                         (slot-value  source 'transforms))))
```
phoe: Why do we use `slot-boundp` and `slot-value` instead of a proper accessor? If I understand the code correctly, the `:initform` of that slot can be `NIL` if we want to have it empty, and this means that we will not have to ever check if it is bound - we just call the accessor, and if the result is non-`NIL`, we proceed.

src/sources/mysql/mysql-schema.lisp
```diff
@@ -4,57 +4,29 @@
...
+    (when (and (slot-boundp source 'fields) (slot-value source 'fields))
+      ;; cast typically happens in copy-database in the schema structure,
+      ;; and the result is then copied into the copy-mysql instance.
+      (unless (and (slot-boundp source 'columns) (slot-value source 'columns))
+        (setf (slot-value source 'columns)
+              (mapcar #'cast (slot-value source 'fields))))
+
+      (unless transforms
```
phoe: The variable `TRANSFORMS` should only be declared here, since we are not using it in the above `WHEN UNLESS SETF` form.

src/sources/mysql/mysql-schema.lisp
```diff
@@ -4,57 +4,29 @@
...
+   (range-list :accessor range-list
+               :initarg :range-list
+               :initform nil))
+  (:documentation "pgloader MySQL Data Source"))
+
+(defmethod initialize-instance :after ((source copy-mysql) &key)
+  "Add a default value for transforms in case it's not been provided."
+  (let ((transforms (and (slot-boundp source 'transforms)
+                         (slot-value  source 'transforms))))
+    (when (and (slot-boundp source 'fields) (slot-value source 'fields))
```
phoe: Ditto.

src/sources/mysql/mysql-schema.lisp
```diff
@@ -4,57 +4,29 @@
...
+      ;; cast typically happens in copy-database in the schema structure,
+      ;; and the result is then copied into the copy-mysql instance.
+      (unless (and (slot-boundp source 'columns) (slot-value source 'columns))
```
phoe: Ditto.

src/sources/mysql/mysql-schema.lisp
```diff
@@ -69,8 +41,10 @@
   "Associate internal table type symbol with what's found in MySQL
   information_schema.tables.table_type column.")
 
-(defun filter-list-to-where-clause (filter-list &optional not)
+(defmethod filter-list-to-where-clause ((mysql copy-mysql) filter-list
+                                        &key not &allow-other-keys)
   "Given an INCLUDING or EXCLUDING clause, turn it into a MySQL WHERE clause."
+  (declare (ignore mysql))
```
phoe: The declaration is not required in a method where you specialize on the argument in question. The compiler should not warn you about this argument not being used since you *already* use it for method dispatch.

src/sources/mysql/mysql-schema.lisp
```diff
@@ -131,23 +105,18 @@
...
-                         &key
-                           only-tables
-                           including
-                           excluding)
+(defmethod fetch-indexes ((schema schema) (mysql copy-mysql)
+                          &key including excluding)
   "Get the list of MySQL index definitions per table."
   (loop
      :for (table-name name index-type non-unique cols)
      :in (mysql-query (format nil
```
phoe: This `FORMAT` call, `(format nil (sql X) Y including including excluding excluding)`, repeats itself many times inside the code. Maybe it is worth to extract it into a helper function?

Even more, the call is like `(mysql-query (format ...))`, `(mssql-query (format ...))`, ... Maybe it would be worth to create a generic function and specialize it on the subclasses of `copy-db` that you have nearby?

src/sources/mysql/mysql-schema.lisp
```diff
@@ -259,42 +216,22 @@
...
-     for is-null = (and next-null (string= next-col "1"))
-     for is-empty = (and next-null (string= next-col "0") (null current-col))
-     ;; don't collect columns we added, e.g. "column_name is not null"
-     when (not current-null)
-     collect (cond (is-null :null)
-		   (is-empty "")
-		   (t current-col))))
-
+  (loop :for field :in (fields mysql)
+     :collect (let ((name (mysql-column-name field))
```
phoe: Nested `LET` inside `LOOP` is not idiomatic. I suggest:

```lisp
(defmethod get-column-list ((mysql copy-mysql))
  "Some MySQL datatypes have a meaningless default output representation, we
   need to process them on the SQL side (geometric data types)."
  (loop :for field :in (fields mysql)
        :for name := (mysql-column-name field)
        :for type := (mysql-column-dtype field)
        :collect (get-column-sql-expression mysql name type)))
```

src/sources/sqlite/sqlite-schema.lisp
```diff
@@ -3,66 +3,60 @@
...
-    (let ((sql (format nil (sql "/sqlite/sqlite-sequence.sql"))))
-      (log-message :sql "SQLite: ~a" sql)
-      (when (sqlite:execute-single (conn-handle slconn) sql)
-        (setf (has-sequences slconn) t))))
-  slconn)
+(defmethod initialize-instance :after ((source copy-sqlite) &key)
+  "Add a default value for transforms in case it's not been provided."
+  (let* ((transforms (when (slot-boundp source 'transforms)
+		       (slot-value source 'transforms))))
+    (when (and (slot-boundp source 'fields) (slot-value source 'fields))
```
phoe: Ditto.

src/sources/sqlite/sqlite-schema.lisp
```diff
@@ -3,66 +3,60 @@
...
-  (when check-has-sequences
-    (let ((sql (format nil (sql "/sqlite/sqlite-sequence.sql"))))
-      (log-message :sql "SQLite: ~a" sql)
-      (when (sqlite:execute-single (conn-handle slconn) sql)
-        (setf (has-sequences slconn) t))))
-  slconn)
+(defmethod initialize-instance :after ((source copy-sqlite) &key)
+  "Add a default value for transforms in case it's not been provided."
+  (let* ((transforms (when (slot-boundp source 'transforms)
+		       (slot-value source 'transforms))))
```
phoe: This line mixes tabs and spaces. Also, ditto.

src/sources/sqlite/sqlite-schema.lisp
```diff
@@ -3,66 +3,60 @@
...
+    (when (and (slot-boundp source 'fields) (slot-value source 'fields))
+      ;; cast typically happens in copy-database in the schema structure,
+      ;; and the result is then copied into the copy-mysql instance.
+      (unless (and (slot-boundp source 'columns) (slot-value source 'columns))
```
phoe: Ditto.

src/sources/mssql/mssql-schema.lisp
```diff
@@ -4,48 +4,26 @@
...
-  (query *mssql-db* query))
+(defclass copy-mssql (db-copy)
+  ((encoding :accessor encoding         ; allows forcing encoding
+             :initarg :encoding
+             :initform nil))
+  (:documentation "pgloader MS SQL Data Source"))
+
+(defmethod initialize-instance :after ((source copy-mssql) &key)
+  "Add a default value for transforms in case it's not been provided."
+  (let* ((transforms (when (slot-boundp source 'transforms)
```
phoe: Same issue as the one I've mentioned below.

src/sources/mssql/mssql-schema.lisp
```diff
@@ -4,48 +4,26 @@
...
+		       (slot-value source 'transforms))))
+    (when (and (slot-boundp source 'fields) (slot-value source 'fields))
```
phoe: Ditto.

src/sources/mssql/mssql-schema.lisp
```diff
@@ -4,48 +4,26 @@
...
+      ;; cast typically happens in copy-database in the schema structure,
+      ;; and the result is then copied into the copy-mysql instance.
+      (unless (and (slot-boundp source 'columns) (slot-value source 'columns))
```
phoe: Ditto.

src/sources/db3/db3-schema.lisp
dimitri: It's my Emacs setup that does that, I think. And I think it's just standard SLIME setup here. In other words I don't mind too much about that.

src/sources/mssql/mssql-connection.lisp
dimitri: Thanks, fixed.

src/sources/mssql/mssql-schema.lisp
dimitri: Let's try and be more idiomatic then! Thanks for the example, I've done it that way now. Only change is that since I've been beaten in the past with binding scopes and overloads in LOOP, I changed the bindings we use at the :with level. The including/excluding clauses from the pgloader command grammar are turned into SQL where clause parts, now named incl-where and excl-where.

src/sources/sqlite/sqlite-schema.lisp
dimitri: It turns out that this code is a vestige from old times. Nowadays the cast function is taken care of at the `process-catalog` level when preparing our work specification, and then the catalog bits are copied over to the COPY objects. I removed the `initialize-instance :after` methods, we don't need them anymore.

src/sources/sqlite/sqlite-schema.lisp
dimitri: Fixed.

src/sources/mssql/mssql-schema.lisp
dimitri: Fixed.

src/sources/mssql/mssql-schema.lisp
dimitri: Fixed.

src/sources/mssql/mssql-schema.lisp
dimitri: Fixed.

src/sources/mssql/mssql-schema.lisp
dimitri: Fixed.

src/sources/mysql/mysql-schema.lisp
dimitri: Good thinking. It's too invasive for this PR though. Also I wonder if I should get rid of the COPY idea and rely more on the catalogs, where most of the work is done nowadays anyway.

src/sources/mysql/mysql-schema.lisp
dimitri: It's old code that not even needed anymore, I just killed it for good. Also it was me still discovering Common Lisp at the times I wrote that, so it's just patching it until it works rather than careful reasoning here.

src/sources/mysql/mysql-schema.lisp
dimitri: Code removed.

src/sources/mysql/mysql-schema.lisp
dimitri: Fixed.

src/sources/mysql/mysql-schema.lisp
dimitri: That's a good point. I did a quick improvement here by moving the FORMAT call into the SQL function. I think it makes sense from an abstraction perspective and would make more sense if we ever got to change our templating engine from FORMAT to something else.

I'll mark this comment/conversation as *resolved* for now, please feel free to open an issue and then a PR to improve the situation even more, or see if using cl-yesql would make sense for pgloader.

dimitri:
> This may signal a type error. `(gethash url *fs*)` may return `NIL` which is not a valid format string. The `OR` clause is therefore badly constructed and the `ERROR` call in line 65 is unreachable.

Thanks, good spot! Fixed now!

PR #986, opened 2019-06-11
Blockstack core pgloader

GInxh:


dimitri:
Is there an intention to contribute something to the pgloader official docker image? that'd be welcome!

GInxh:
no, the default selection for creating a PR request when forked in a private repo is the public one. However, none of this data is sensitive. The owner of the repo (you) can delete this if you desire. It is up to you in the name of transparency. None of this was intended to be used for any development of pgloader itself. 

The patch file was edited to use inotify on migration, but it turns out inotify watch flags do not operate on the final output desribed in a binary fashion, but underlying actions from the system calls that do not work when a file moves across nodes on a multi kernel architecture, so most of this implementation is deprecated in our final merge anyways.

Issue #1008, opened 2019-08-06
Undefined foreign library: CL+SSL::LIBEAY32


mrohnstock:
- [ ] pgloader --version
  
    ```
    3.6.1
    ```
  
  - [ ] did you test a fresh compile from the source tree?
    
    yes - tag 3.6.1
    
  - [ ] did you search for other similar issues?
    
    yes, but can only find issues with windows, not linux.

  - [ ] how can I reproduce the bug?

    I'm using the AUR-Package from [here](https://aur.archlinux.org/packages/pgloader) on my archlinux system. Building the package worked without issues, but executing the following (or any other) command:

```
pgloader --version
```

the following messages occur:

```
debugger invoked on a SIMPLE-ERROR in thread
#<THREAD "main thread" RUNNING {10006405B3}>:
  Problem running initialization hook #<FUNCTION OPEN-FOREIGN-LIBS>:
  Undefined foreign library: CL+SSL::LIBEAY32

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [CONTINUE] Skip this initialization hook.
  1: [ABORT   ] Exit from the current thread.

(SB-INT:CALL-HOOKS "initialization" (#<FUNCTION OPEN-FOREIGN-LIBS>) :ON-ERROR :ERROR)
0]
```

I think a dependency is missing, but I can't find the missing one - as compiling worked without issues.

Thanks :)

rusq:
facing the same error when trying to start after building in docker:

Sample docker file:
```docker
ARG DYNSIZE=2048

FROM ubuntu:xenial

ARG DYNSIZE

RUN apt-get update 
RUN apt-get install -y sbcl unzip libsqlite3-dev make curl gawk freetds-dev libzip-dev libssl-dev
RUN apt-get install -y git

WORKDIR /opt

RUN git clone https://github.com/dimitri/pgloader

WORKDIR /opt/pgloader

RUN git checkout v3.6.1

RUN make DYNSIZE=${DYNSIZE} pgloader
```

Error when trying to start:
```
root@7933eb8d3da5:/opt/pgloader/build/bin# ./pgloader

debugger invoked on a SIMPLE-ERROR in thread
#<THREAD "main thread" RUNNING {1009DBC743}>:
  Problem running initialization hook #<FUNCTION OPEN-FOREIGN-LIBS>:
[...]

PR #1010, opened 2019-08-12
Add missing cl-plus-ssl dep

rusq:
Fixes issue #1008


rusq:
@dimitri man, you really need to fix your travis build scripts.

PR #1027, opened 2019-09-23
add big json support to pgsql

luongvm:
When loading from mysql -> s3 -> redshift, I encountered a problem: json fields with big length create error because the default VARCHAR(512) is not enough.

This is the simplest fix I can think of, if anyone have better suggestion, that doesn't involve changing the source code please let me know and I will adjust the code.


dimitri:
Large JSON support in pgsql→pgsql migrations. The v3 codebase is in maintenance mode. **pgloader v4** uses JDBC streaming and does not have the same size constraints. If you still hit issues with v4, please open a new issue.

---

PR #1083, opened 2020-02-20
Disable attempts to drop constraints

AD7six:


src/pgsql/pgsql-ddl.lisp
```diff
@@ -306,7 +306,7 @@
 (defmethod format-drop-sql ((fk fkey) &key (stream nil) cascade if-exists)
   (let* ((constraint-name (fkey-name fk))
          (table-name      (format-table-name (fkey-table fk))))
-    (format stream "ALTER TABLE ~a DROP CONSTRAINT~:[~; IF EXISTS~] ~a~@[ CASCADE~];"
+    (format stream "disable_foreign_key_constraints()"
```
AD7six: This'll get called more than once which is suboptimal, but a viable workaround to running with a none-superuser account

AD7six:
Sorry this was not intended to be a PR to the origin repo :D

PR #1107, opened 2020-03-22
fix typo that chooses wrong option rule

svantevonerichsen6906:
Just spotted


dimitri:
Thanks @svantevonerichsen6906!

PR #1110, opened 2020-03-24
Change Travis build matrix

### LINKED ISSUE #1108: Travis only builds pgloader against postgres 9.6
https://travis-ci.org/github/dimitri/pgloader/builds/665623197

I can see that the two passing jobs explicitly set Postgres version to 9.6. Shouldn't we also test against versions 10, 11, and 12, since these versions are available and supported?

### LINKED ISSUE #1109: Travis job fails if PGVERSION is not set
https://travis-ci.org/github/dimitri/pgloader/jobs/665623198#L228

```
$ ./.travis.sh postgresql_install
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
The command "./.travis.sh postgresql_install" failed and exited with 2 during .
```

I assume that this fails due to:
https://github.com/dimitri/pgloader/blob/781e58681624adebecc15a9416b718c2023c8fec/.travis.sh#L45-L48
There is no `psql` command to execute to get the current postgres version from.

Should this test configuration be kept in Travis?

phoe:
We remove the Travis jobs without the `PGLOADER` environment variable set (to try and fix #1109) and add Postgres versions 10, 11, and 12 to build and test pgloader against them (to try and fix #1108).

Please merge only if [the related Travis job](https://travis-ci.org/github/dimitri/pgloader/builds/666266112) succeeds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,14 @@ sudo: required
 
 env:
   matrix:
-    - LISP=ccl
     - LISP=ccl  PGVERSION=9.6
-    - LISP=sbcl
+    - LISP=ccl  PGVERSION=10
+    - LISP=ccl  PGVERSION=11
+    - LISP=ccl  PGVERSION=12
     - LISP=sbcl PGVERSION=9.6
+    - LISP=sbcl PGVERSION=10
+    - LISP=sbcl PGVERSION=11
+    - LISP=sbcl PGVERSION=12
 
 install:
   - ./.travis.sh lisp_install


phoe:
Zrzut ekranu z 2020-03-24 11-34-18

*(Japanese mahjong has a term for such a situation - it's 綠一色, also known as [All Green](http://mahjong.wikidot.com/all-green). A pretty precious hand to assemble.)*

Is this enough to merge the PR and get meaningfully green Travis builds once more, or is there some functionality that we lose by dropping the `PGVERSION`-less builds? (I don't think there is, since the jobs were failing so early in the build process, but it's best to ask.)

phoe:
OK - will do that tomorrow.

PR #1112, opened 2020-03-24
Simplify and fix errors/warnings/notes on Travis configuration

### LINKED ISSUE #1111: Travis configuration errors/warnings/notes
https://travis-ci.org/github/dimitri/pgloader/jobs/666266506/config

> **Build config validation**
> * :x: language: unknown value common-lisp
> * :x: language: unknown value common-lisp
> * :warning: root: deprecated key sudo (The key `sudo` has no effect anymore.)
> * :warning: root: deprecated key sudo (The key `sudo` has no effect anymore.)
> * :heavy_check_mark: env: key matrix is an alias for jobs, using jobs
> * :heavy_check_mark: root: missing dist, using the default xenial
> * :heavy_check_mark: root: missing os, using the default linux
> * :heavy_check_mark: env: key matrix is an alias for jobs, using jobs 

I think it would be worth to clean these up.

Language can be set to `shell` or `generic`, as according to [the Travis documentation](https://docs.travis-ci.com/user/languages/minimal-and-generic/); `sudo` can be removed; `matrix` substituted for `jobs`; `dist` and `os` can be set explicitly.

phoe:
Fixes #1111.

We remove unnecessary .travis.sh code for removing old Postgres versions.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -43,20 +43,9 @@ pgdg_repositories() {
 }
 
 postgresql_install() {
-	if [ -z "${PGVERSION:-}" ]; then
-		PGVERSION="$( psql -d postgres -XAtc "select regexp_replace(current_setting('server_version'), '[.][0-9]+$', '')" )"
-	else
-		sudo service postgresql stop
-		xargs sudo apt-get -y --purge remove <<-packages
-			libpq-dev
-			libpq5
-			postgresql
-			postgresql-client-common
-			postgresql-common
-		packages
-		sudo rm -rf /var/lib/postgresql
-	fi
-
+        if [ -z "${PGVERSION:-}" ]; then
+                PGVERSION="$( psql -d postgres -XAtc "select regexp_replace(current_setting('server_version'), '[.][0-9]+$', '')" )"
+        fi
 	xargs sudo apt-get -y install <<-packages
 		postgresql-${PGVERSION}
 		postgresql-${PGVERSION}-ip4r
--- a/.travis.sh
+++ b/.travis.sh
@@ -43,9 +43,10 @@ pgdg_repositories() {
 }
 
 postgresql_install() {
-        if [ -z "${PGVERSION:-}" ]; then
-                PGVERSION="$( psql -d postgres -XAtc "select regexp_replace(current_setting('server_version'), '[.][0-9]+$', '')" )"
-        fi
+	if [ -z "${PGVERSION:-}" ]; then
+		echo 'PGVERSION environment variable not set.';
+		exit 1
+	fi
 	xargs sudo apt-get -y install <<-packages
 		postgresql-${PGVERSION}
 		postgresql-${PGVERSION}-ip4r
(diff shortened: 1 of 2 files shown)


.travis.sh
```diff
@@ -43,20 +43,9 @@ pgdg_repositories() {
...
-			libpq5
-			postgresql
-			postgresql-client-common
-			postgresql-common
-		packages
-		sudo rm -rf /var/lib/postgresql
-	fi
-
+        if [ -z "${PGVERSION:-}" ]; then
+                PGVERSION="$( psql -d postgres -XAtc "select regexp_replace(current_setting('server_version'), '[.][0-9]+$', '')" )"
```
dimitri: There's a simpler way to do it:

```
psql -At -c "select current_setting('server_version_num')::int / 10000"
```

How to get there:

```
pgloader# select version();
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                   version                                                    │
├──────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ PostgreSQL 10.10 on x86_64-apple-darwin14.5.0, compiled by Apple LLVM version 7.0.0 (clang-700.1.76), 64-bit │
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(1 row)

pgloader# select name, setting from pg_settings where name ~ 'version';
┌────────────────────┬─────────┐
│        name        │ setting │
├────────────────────┼─────────┤
│ server_version     │ 10.10   │
│ server_version_num │ 100010  │
└────────────────────┴─────────┘
(2 rows)

pgloader# select current_setting('server_version_num')::int / 10000;
┌──────────┐
│ ?column? │
├──────────┤
│       10 │
└──────────┘
(1 row)
```

phoe: OK, will do that before merging. Thanks for the comment!

phoe: @dimitri Question - now that the buildmatrix only has jobs that explicitly specify `PGVERSION`, is there any sense in keeping this whole codeblock? It's effectively dead code now that it doesn't run in Travis.

```bash
        if [ -z "${PGVERSION:-}" ]; then
                PGVERSION="$( psql -At -c "select current_setting('server_version_num')::int / 10000" )"
        fi
```

dimitri: Oh dead code... then we should remove it I think, thanks for this maintenance efforts, much appreciated!

phoe: OK - the current version of my PR has that code removed.

I am ready to squash&merge!

dimitri: If you have more time to spend on our tests I would suggest to have a look at the src/regress tooling in the Lisp code and see how to support database sources rather than single-table sources, so that we could run SQLite and MySQL migration tests in Travis too...

That's for another PR, of course, and would be great to have that. Another of the things I have in the back of my mind and never got around to...

phoe: Created an issue for that.

Should I merge this one up?

dimitri: Yeah looks like it's ready!

PR #1120, opened 2020-03-28
Implement tables row count ordering for MySQL.

### LINKED ISSUE #1099: [Feature request] Order source tables by size descending to reduce migration time
Hi,

first thank you for this great software! I am used to write my migration scripts by hand for other destination DBs and pgloader is really wonderful!

From previous migrations I have the experience to sort the tables by size descending, so the biggest tables in a parallel migration are started first.

By using this strategy the runtime of the migration is in most cases the time for the biggest table - more so if there is one table outstanding in size.

I am currently migrating a zabbix DB from mysql to postgresql, the history and trends tables are huge compared to the rest (260GByte vs. 2GByte).

All the best and thanks!
Rainer

dimitri:
This should help optimize the duration of migrating databases with very big
tables and lots of smaller ones. It might be a little too naive as far as
the optimisation goes, while still being an improvement on the default
alphabetical one.

Fixes #1099.

### DIFF
--- a/src/load/migrate-database.lisp
+++ b/src/load/migrate-database.lisp
@@ -255,6 +255,29 @@
     (setf (catalog-distribution-rules catalog)
           (citus-distribute-schema catalog distribute))))
 
+(defun optimize-table-copy-ordering (catalog)
+  "Return a list of tables to copy over in optimized order"
+  (let ((table-list (copy-list (table-list catalog)))
+        (view-list  (copy-list (view-list catalog))))
+    ;; when materialized views are not supported, view-list is empty here
+    (cond
+      ((notevery #'null (mapcar #'table-row-count-estimate table-list))
+           (flet ((compare-row-count-estimates (a b)
+                    (declare (type (or null fixnum) a b))
+                    (> (or a 0) (or b 0))))
+             (let ((sorted-table-list (sort table-list
+                                            #'compare-row-count-estimates
+                                            :key #'table-row-count-estimate)))
+               (log-message :notice
+                            "Processing tables in this order: ~{~a: ~d rows~^, ~}"
+                            (loop :for table
+                               :in (append sorted-table-list view-list)
+                               :append (list (format-table-name table)
+                                             (table-row-count-estimate table))))
+               (append sorted-table-list view-list))))
+      (t
+       (append table-list view-list)))))
+
 
 ;;;
 ;;; Generic enough implementation of the copy-database method.
@@ -414,10 +437,7 @@
         (return-from copy-database)))
 
     (loop
-       :for table :in (append (table-list catalog)
-                              ;; when materialized views are not supported,
-                              ;; view-list is empty here
-                              (view-list catalog))
+       :for table :in (optimize-table-copy-ordering catalog)
 
        :do (let ((table-source (instanciate-table-copy-object copy table)))
              ;; first COPY the data from source to PostgreSQL, using copy-kernel
@@ -472,8 +492,21 @@
 
                      (when (and create-indexes
                                 (zerop (gethash table writers-count)))
-                       (log-message :notice "DONE copying ~a"
-                                    (format-table-name table))
+
+                       (let* ((stats   pgloader.monitor::*sections*)
+                              (section (get-state-section stats :data))
+                              (table-stats (pgstate-get-label section table))
+                              (pprint-secs
+                               (pgloader.state::format-interval seconds nil)))
+                         ;; in CCL we have access to the *sections* dynamic
+                         ;; binding from another thread, in SBCL we access
+                         ;; an empty copy.
+                         (log-message :notice
+                                      "DONE copying ~a in ~a~@[ for ~d rows~]"
+                                     (format-table-name table)
+                                     pprint-secs
+                                     (when table-stats
+                                       (pgtable-rows table-stats))))
                        (alexandria:appendf
                         pkeys
                         (create-indexes-in-kernel (target-db copy)
--- a/src/load/migrate-database.lisp
+++ b/src/load/migrate-database.lisp
@@ -261,22 +261,17 @@
         (view-list  (copy-list (view-list catalog))))
     ;; when materialized views are not supported, view-list is empty here
     (cond
-      ((notevery #'null (mapcar #'table-row-count-estimate table-list))
-           (flet ((compare-row-count-estimates (a b)
-                    (declare (type (or null fixnum) a b))
-                    (> (or a 0) (or b 0))))
-             (let ((sorted-table-list (sort table-list
-                                            #'compare-row-count-estimates
-                                            :key #'table-row-count-estimate)))
-               (log-message :notice
-                            "Processing tables in this order: ~{~a: ~d rows~^, ~}"
-                            (loop :for table
-                               :in (append sorted-table-list view-list)
-                               :append (list (format-table-name table)
-                                             (table-row-count-estimate table))))
-               (append sorted-table-list view-list))))
+      ((notevery #'zerop (mapcar #'table-row-count-estimate table-list))
+       (let ((sorted-table-list
+              (sort table-list #'> :key #'table-row-count-estimate)))
+         (log-message :notice
+                      "Processing tables in this order: ~{~a: ~d rows~^, ~}"
+                      (loop :for table :in (append table-list view-list)
+                         :collect (format-table-name table)
+                         :collect (table-row-count-estimate table)))
+         (nconc sorted-table-list view-list)))
       (t
-       (append table-list view-list)))))
+       (nconc table-list view-list)))))
 
 
 ;;;
--- a/src/sources/common/api.lisp
+++ b/src/sources/common/api.lisp
@@ -153,6 +153,9 @@
 (defgeneric fetch-foreign-keys (catalog db-copy &key including excluding)
   (:documentation "Get the list of foreign keys from the source database."))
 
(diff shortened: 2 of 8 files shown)


src/load/migrate-database.lisp
```diff
@@ -472,8 +492,21 @@
...
-                                    (format-table-name table))
+
+                       (let* ((stats   pgloader.monitor::*sections*)
+                              (section (get-state-section stats :data))
+                              (table-stats (pgstate-get-label section table))
+                              (pprint-secs
+                               (pgloader.state::format-interval seconds nil)))
+                         ;; in CCL we have access to the *sections* dynamic
+                         ;; binding from another thread, in SBCL we access
+                         ;; an empty copy.
```
phoe: Do we need a singleton object accessible from all threads in order to access some monitor data? If that's the case, we can use https://github.com/lmj/global-vars that is well-suited for that use case.

src/load/migrate-database.lisp
```diff
@@ -255,6 +255,29 @@
...
+                    (> (or a 0) (or b 0))))
+             (let ((sorted-table-list (sort table-list
+                                            #'compare-row-count-estimates
+                                            :key #'table-row-count-estimate)))
+               (log-message :notice
+                            "Processing tables in this order: ~{~a: ~d rows~^, ~}"
+                            (loop :for table
+                               :in (append sorted-table-list view-list)
+                               :append (list (format-table-name table)
+                                             (table-row-count-estimate table))))
```
phoe: `:append (list ...)` conses much; I'd instead suggest:

```lisp
(loop ...
  :collect (format-table-name table)
  :collect (table-row-count-estimate-table))
```

src/load/migrate-database.lisp
```diff
@@ -255,6 +255,29 @@
...
+           (flet ((compare-row-count-estimates (a b)
+                    (declare (type (or null fixnum) a b))
+                    (> (or a 0) (or b 0))))
+             (let ((sorted-table-list (sort table-list
+                                            #'compare-row-count-estimates
+                                            :key #'table-row-count-estimate)))
+               (log-message :notice
+                            "Processing tables in this order: ~{~a: ~d rows~^, ~}"
+                            (loop :for table
+                               :in (append sorted-table-list view-list)
```
phoe: `sorted-table-list` is fresh and straight out of `sort`; we may use `nconc` instead of `append` in this line.

src/load/migrate-database.lisp
```diff
@@ -255,6 +255,29 @@
...
+(defun optimize-table-copy-ordering (catalog)
+  "Return a list of tables to copy over in optimized order"
+  (let ((table-list (copy-list (table-list catalog)))
+        (view-list  (copy-list (view-list catalog))))
+    ;; when materialized views are not supported, view-list is empty here
+    (cond
+      ((notevery #'null (mapcar #'table-row-count-estimate table-list))
+           (flet ((compare-row-count-estimates (a b)
+                    (declare (type (or null fixnum) a b))
+                    (> (or a 0) (or b 0))))
```
phoe: Accessing uninitialized slots here. See the comment on `DEFSTRUCT TABLE`.

src/utils/catalog.lisp
```diff
@@ -48,11 +48,16 @@
            table-list view-list matview-list extension-list sqltype-list)
 
 (defstruct table source-name name schema oid comment
-           storage-parameter-list tablespace
+           storage-parameter-list tablespace row-count-estimate
```
phoe: Undefined behaviour.

In `optimize-table-copy-ordering` in `migrate-database.lisp` we are accessing the value of slot `row-count-estimate` in structure `table` without setting it first.

[CLHS DEFSTRUCT](http://clhs.lisp.se/Body/m_defstr.htm) states:
> If no `slot-initform` is supplied, the consequences are undefined if an attempt is later made to read the slot's value before a value is explicitly assigned.

We are lucky that SBCL and CCL initialize these values to `NIL`, but they are not required to.

The slot definition `row-count-estimate` should be changed to `(row-count-estimate nil)` to explicitly initialize the slot if we want to read it without setting. Orrrr, we can even initialize it to `0` if that is the value that we are looking for in `optimize-table-copy-ordering`; that will simplify the `FLET` found there into:
```lisp
           (flet ((compare-row-count-estimates (a b)
                    (declare (fixnum a b))
                    (> a b)))
```
At this point, the function becomes unnecessary, since it becomes just a `#'>` call.

src/load/migrate-database.lisp
```diff
@@ -255,6 +255,29 @@
...
+                                            :key #'table-row-count-estimate)))
+               (log-message :notice
+                            "Processing tables in this order: ~{~a: ~d rows~^, ~}"
+                            (loop :for table
+                               :in (append sorted-table-list view-list)
+                               :append (list (format-table-name table)
+                                             (table-row-count-estimate table))))
+               (append sorted-table-list view-list))))
+      (t
+       (append table-list view-list)))))
```
phoe: `table-list` and `view-list` are both fresh; we can use `nconc` instead of `append` here.

src/load/migrate-database.lisp
```diff
@@ -255,6 +255,29 @@
     (setf (catalog-distribution-rules catalog)
           (citus-distribute-schema catalog distribute))))
 
+(defun optimize-table-copy-ordering (catalog)
+  "Return a list of tables to copy over in optimized order"
+  (let ((table-list (copy-list (table-list catalog)))
+        (view-list  (copy-list (view-list catalog))))
+    ;; when materialized views are not supported, view-list is empty here
+    (cond
+      ((notevery #'null (mapcar #'table-row-count-estimate table-list))
```
phoe: Accessing uninitialized slots here. See the comment on `DEFSTRUCT TABLE`.

src/load/migrate-database.lisp
```diff
@@ -255,6 +255,29 @@
...
+             (let ((sorted-table-list (sort table-list
+                                            #'compare-row-count-estimates
+                                            :key #'table-row-count-estimate)))
+               (log-message :notice
+                            "Processing tables in this order: ~{~a: ~d rows~^, ~}"
+                            (loop :for table
+                               :in (append sorted-table-list view-list)
+                               :append (list (format-table-name table)
+                                             (table-row-count-estimate table))))
+               (append sorted-table-list view-list))))
```
phoe: `sorted-table-list` and `view-list` are both fresh; we can use `nconc` instead of `append` here.

src/sources/mysql/mysql-schema.lisp
```diff
@@ -155,6 +155,27 @@
...
 
+;;;
+;;; MySQL table row count estimate
+;;;
+(defmethod fetch-table-row-count ((schema schema)
+                                  (mysql copy-mysql)
+                                  &key
+                                    including
+                                    excluding)
+  "Get the row count estimate for given MySQL tables."
```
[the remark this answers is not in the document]
dimitri: Good suggestion, applied, thanks!

src/sources/common/api.lisp
```diff
@@ -153,6 +153,9 @@
 (defgeneric fetch-foreign-keys (catalog db-copy &key including excluding)
   (:documentation "Get the list of foreign keys from the source database."))
 
+(defgeneric fetch-table-row-count (catalog db-copy &key including excluding)
+  (:documentation "Get the estimate row count for tables in the source database"))
```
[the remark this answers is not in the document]
dimitri: Done.

src/load/migrate-database.lisp
dimitri: That's good thinking, and I think that's relevant for another PR. The PR introducing the statistics singleton should also take care of the following bits of refactoring of the code:

  - the singleton is named “state” as I initially though I would update the global state of the  migration to be able to know about the progress being made while it's being made,
  - when implementing the feature, it quickly enough turned into statistics (count of rows read, imported, errors, timings, etc) rather “state”
  - the API to access and maintain the state objects (stats really) per table etc is rather lacking, so we should clean that up and offer easy access to needed information from threads

At the moment only the monitor thread is creating and updating the data, we certainly want to keep it that way. That allows no-locks (and no dead locks) code, with using a queue to send events and centralise statistics from.

src/load/migrate-database.lisp
dimitri: Done, thanks for the hint!

src/load/migrate-database.lisp
dimitri: Done.

Note that pgloader makes almost no efficiency efforts when dealing with the catalogs, because that's not where the program spend its time on when running anyway. It's all about moving the data. [Amdahl's law](https://en.wikipedia.org/wiki/Amdahl%27s_law) would tell us that the best we can optimise for in that area of the code might be in the are of 1% to 5% of the run-time, I would guess.

src/load/migrate-database.lisp
dimitri: Good point, now added a default value of zero and a type for this slot in the structure.

src/utils/catalog.lisp
dimitri: It's all simplified down thanks to better type definition, thanks!

src/load/migrate-database.lisp
dimitri: Done.

src/load/migrate-database.lisp
dimitri: Done.

Issue #1127, opened 2020-04-07
ERROR mysql: The value NIL is not of type STRING when binding STRING


gavinwahl:
- pgloader --version
  
    ```
    pgloader version "3.6.86b6a5c"
    compiled with SBCL 1.4.14-3.fc31
    ```
  
  - did you test a fresh compile from the source tree?
    Yes, tested this on master commit 86b6a5cb8

```
$ ./build/bin/pgloader --verbose --debug mysql://root@localhost/XXX pgsql:///XXX
pgloader version 3.6.86b6a5c
compiled with SBCL 1.4.14-3.fc31
sb-impl::*default-external-format* :UTF-8
tmpdir: #P"/tmp/pgloader/"
2020-04-06T18:13:23.022000-06:00 NOTICE Starting pgloader, log system is ready.
2020-04-06T18:13:23.043000-06:00 LOG pgloader version "3.6.86b6a5c"
2020-04-06T18:13:23.044000-06:00 DEBUG LOAD DATA FROM #<PGLOADER.SOURCE.MYSQL:MYSQL-CONNECTION mysql://root@localhost:3306/XXX {100560FA33}>
2020-04-06T18:13:23.074000-06:00 DEBUG CONNECTED TO #<PGLOADER.PGSQL:PGSQL-CONNECTION pgsql://gavin@UNIX:5432/XXX {10058137C3}>
2020-04-06T18:13:23.074000-06:00 DEBUG SET client_encoding TO 'utf8'
2020-04-06T18:13:23.074000-06:00 DEBUG SET application_name TO 'pgloader'
2020-04-06T18:13:23.092000-06:00 LOG Migrating from #<MYSQL-CONNECTION mysql://root@localhost:3306/XXX {100560FA33}>
2020-04-06T18:13:23.092000-06:00 LOG Migrating into #<PGSQL-CONNECTION pgsql://gavin@UNIX:5432/XXX {10058137C3}>
2020-04-06T18:13:23.148000-06:00 DEBUG CONNECTED TO #<MYSQL-CONNECTION mysql://root@localhost:3306/XXX {100560FA33}>
2020-04-06T18:13:23.156000-06:00 SQL MySQL: sending query: -- params: db-name
--         table-type-name
--         only-tables
--         only-tables
--         including
--         filter-list-to-where-clause incuding
--         excluding
--         filter-list-to-where-clause excluding
  select c.table_name, t.table_comment,
         c.column_name, c.column_comment,
[...]

PR #1128, opened 2020-04-07
Fix syntax error near `integer`, and value NIL 

MrSaints:
Since `pgloader` @ 11970bbc, importing from MySQL 5.7 to Postgresql 11
with SSL triggers the following exception:

```
 You have an error in your SQL syntax; check the manual that corresponds
 to your MySQL server version for the right syntax to use near 'integer'
```

Rather than casting to `integer`, we should be casting to `unsigned`
or calling `floor(..)`.

### DIFF
--- a/src/sources/mysql/sql/list-table-rows.sql
+++ b/src/sources/mysql/sql/list-table-rows.sql
@@ -4,7 +4,7 @@
 --         excluding
 --         filter-list-to-where-clause excluding
     SELECT table_name,
-           cast(data_length/avg_row_length as integer)
+           cast(data_length/avg_row_length as unsigned)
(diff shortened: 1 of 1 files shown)


dimitri:
Can you see #1127 and add a COALESCE to your query so that empty tables with a NULL result there are covered too? or maybe that's not necessary with your version of the SQL query?

MrSaints:
I can try that, but I actually tried my version too, and it appears to work fine with it versus without it.

MrSaints:
I've pushed a fix for #1127 as well @dimitri.

I tested it without it, and it breaks when you have an empty table. But, with the `COALESCE`, and the fix to the cast, it all works.

Issue #1144, opened 2020-05-07
MSSQL "quote identifiers" setting is not being applied for foreign keys


semenmiroshnichenko:
- [X] pgloader --version
  
    ```
    3.6.38a62a7
    ```
  
  - [X] did you test a fresh compile from the source tree?
    
    Yes, it's build from sources
    
  - [X] did you search for other similar issues?
    
  - [X] how can I reproduce the bug?

    Create database and tables with following script:

```
CREATE DATABASE [smmigtest2]
GO

USE [smmigtest2]
GO


CREATE TABLE [dbo].[Parent](
	[Id] [uniqueidentifier] NOT NULL,
	[Name] [nvarchar](50) NULL,
 CONSTRAINT [PK_Parent] PRIMARY KEY CLUSTERED 
 (
	[Id] ASC
 )
) ON [PRIMARY]
GO



CREATE TABLE [dbo].[Child](
	[Id] [uniqueidentifier] NOT NULL,
	[Value] [int] NULL,
	[ParentId] [uniqueidentifier] NULL,
 CONSTRAINT [PK_Child] PRIMARY KEY CLUSTERED 
(
	[Id] ASC
)
) ON [PRIMARY]
GO

ALTER TABLE [dbo].[Child]  WITH CHECK ADD  CONSTRAINT [FK_Child_Parent] FOREIGN KEY([Id])
REFERENCES [dbo].[Parent] ([Id])
GO

ALTER TABLE [dbo].[Child] CHECK CONSTRAINT [FK_Child_Parent]
GO

```

Keep in mind the casing of the foreign key FK_Child_Parent.

Then run pgloader as `/build/bin/pgloader --no-ssl-cert-verification -v -d ../../pgloaderstring`

```
load database

    from mssql://USERNAME:PASSWORD@HOST/DATABASE
    into postgresql://USERNAME:PASSWORD@HOST/DATABASE?sslmode=prefer

cast type datetimeoffset to timestamptz
cast type datetime2 to timestamptz
cast type uniqueidentifier to uuid using sql-server-uniqueidentifier-to-uuid

with prefetch rows=5000, batch rows=1000, quote identifiers, create indexes, include drop, create tables

set work_mem to '16MB', maintenance_work_mem to '512 MB'

before load do $$ drop schema if exists dbo cascade; $$;

```

  - [X] pgloader output you obtain
  
```
[...]

PR #1145, opened 2020-05-07
Keep name casing for foreign keys on mssql

semenmiroshnichenko:
It solves https://github.com/dimitri/pgloader/issues/1144

### DIFF
--- a/src/sources/mssql/mssql-schema.lisp
+++ b/src/sources/mssql/mssql-schema.lisp
@@ -146,7 +146,7 @@
                 (col-name  (apply-identifier-case col))
                 (fcol-name (apply-identifier-case fcol))
                 (pg-fkey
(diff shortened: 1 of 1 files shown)


dimitri:
Thanks @semenmiroshnichenko ! Would you be interested in a contributor access to pgloader repository on GitHub? We sometimes have more MS SQL bugs or improvements to make, and having someone using MS SQL in the team would then be very valuable!

The idea if you submit changes would be to continue using Pull Requests and have them reviewed, of course.

semenmiroshnichenko:
Welcome @dimitri ! Thank you for asking, and I could be able to fix problems we are facing during our migration project (I'm not a lisp guy), for sure I will create PR's with such fixes, but at the moment I'm afraid I will not be able to spend more time at the pgloader.

PR #1179, opened 2020-07-06
Upgrades debian version in docker image

### LINKED ISSUE #1178: SBCL Docker image not being generated
- [ ] pgloader --version
  
    ```
    3.6.2
    ```
  
  - [ ] did you test a fresh compile from the source tree?
    
    No. Tried docker build . -t pgloader
    
  - [ ] did you search for other similar issues?
    
  - [ ] how can I reproduce the bug?
git checkout tags/v3.6.2 -b 3.6.2 && docker build . -t pgloader

Error:
[package zs3]............................
To load "pgloader":
  Load 1 ASDF system:
    pgloader
; Loading "pgloader"
While evaluating the form starting at line 29, column 0
  of #P"/opt/src/pgloader/./src/save.lisp":

debugger invoked on a SB-INT:STREAM-DECODING-ERROR in thread
#<THREAD "main thread" RUNNING {10005D05B3}>:
  :ASCII stream decoding error on
  #<SB-SYS:FD-STREAM for "file /opt/src/pgloader/build/quicklisp/dists/quicklisp/software/uax-15-20200325-git/data/CompositionExclusions.txt" {10028F6D63}>:

    the octet sequence #(194) cannot be decoded.

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [ATTEMPT-RESYNC               ] Attempt to resync the stream at a
                                     character boundary and continue.
  1: [FORCE-END-OF-FILE            ] Force an end of file.
  2: [INPUT-REPLACEMENT            ] Use string as replacement input, attempt
                                     to resync at a character boundary and
                                     continue.
  3: [TRY-RECOMPILING              ] Recompile precomputed-tables and try loading it again
  4: [RETRY                        ] Retry
                                     loading FASL for #<CL-SOURCE-FILE "uax-15" "src" "precomputed-tables">.
  5: [ACCEPT                       ] Continue, treating
                                     loading FASL for #<CL-SOURCE-FILE "uax-15" "src" "precomputed-tables">
                                     as having been successful.
  6:                                 Retry ASDF operation.
  7: [CLEAR-CONFIGURATION-AND-RETRY] Retry ASDF operation after resetting the
                                     configuration.
  8:                                 Retry ASDF operation.
  9:                                 Retry ASDF operation after resetting the
                                     configuration.
 10: [ABORT                        ] Give up on "pgloader"
 11: [RETRY                        ] Retry EVAL of current toplevel form.
 12: [CONTINUE                     ] Ignore error and continue loading file "/opt/src/pgloader/./src/save.lisp".
 13:                                 Abort loading file "/opt/src/pgloader/./src/save.lisp".
 14:                                 Ignore runtime option --load "./src/save.lisp".
 15:                                 Skip rest of --eval and --load options.
 16:                                 Skip to toplevel READ/EVAL/PRINT loop.
 17: [EXIT                         ] Exit SBCL (calling #'EXIT, killing the process).

(SB-IMPL::STREAM-DECODING-ERROR-AND-HANDLE #<SB-SYS:FD-STREAM for "file /opt/src/pgloader/build/quicklisp/dists/quicklisp/software/uax-15-20200325-git/data/CompositionExclusions.txt" {10028F6D63}> 1)
0] 
; 
; compilation unit aborted
;   caught 1 fatal ERROR condition

debugger invoked on a SB-C::INPUT-ERROR-IN-LOAD in thread
#<THREAD "main thread" RUNNING {10005D05B3}>:
  READ error during LOAD:

    Package PGLOADER does not exist.

      Line: 70, Column: 19, File-Position: 2453

      Stream: #<SB-INT:FORM-TRACKING-STREAM for "file /opt/src/pgloader/src/save.lisp" {1001B3B0B3}>

caduvieira:
Upgraded debian docker image for builder and docker final image.

Closes #1178

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stable-slim as builder
+FROM debian:bullseye-slim as builder
 
   RUN apt-get update \
       && apt-get install -y --no-install-recommends \
@@ -28,7 +28,7 @@ FROM debian:stable-slim as builder
       && cd /opt/src/pgloader \
       && make clones save
 
-FROM debian:stable-slim
+FROM debian:bullseye-slim
 
   RUN apt-get update \
       && apt-get install -y --no-install-recommends \


dimitri:
I think we should avoid using static names, and prefer aliases such as stable, testing, and sid. By the way, why using testing here rather than sid? I understand that stable is having problems with some packages versions we depend on (which by the way), but testing sounds like the worst trade-off possible...

phoe:
testing contains packages staged for inclusion into stable and undergoes more rigorous testing than unstable, whereas unstable is allowed to break every now and then.

by the way, "sid" is a static name; did you mean "unstable"? ;)

dimitri:
Well security fixes go straight in sid/unstable. Testing needs to wait for 2 to 10 days before a package that made it to unstable is migrated to testing. Those 10 days involve no extra testing at all. Only if a bug is opened with a strong enough severity then the package migration is put on hold.

Really, I don't think we have any advantage of using testing in that case. Maybe @df7cb could help us make a decision here?

df7cb:
I have no stakes in Docker, but if it helps, I can provide a list of packages that need to be upgraded in buster to build pgloader there. In fact, if you add apt.postgresql.org buster-pgdg, it will contain everything needed. (It will also contain pgloader, so building would not be necessary. :)

If the sbcl version in buster is the problem, we need more regression tests in the Debian package. Last time I tried the regression tests needed a maze of prerequisites to run that I didn't manage to solve, so I went with one single test only (allcols.load).

rodo:
I did a try with ubuntu in the spirit of using aliases instead of static name, I'm joining Dimitri's opinion on this.
Unfortuately the sbcl version in Ubuntu:latest 20.04 today is 2.0.1 with has same problem is does not work, a solution is to use ubuntu:groovy which works but we have the same issue on using static name. By the way the image size for debian:testing is 191MB when ubuntu:groovy is 195MB, quite the same.

dimitri:
In the spirit of having something that just works, what about we merge this PR then, using `bullseye`, and revisit the issue later?

spantaleev:
It's unfortunate that this hasn't been merged for such a long time and that we all have to apply the patch manually when building the image.

-----------

What's wrong with using static names for the base image?

Using something like `stable` seems dangerous, because it's a moving target. It's one thing today, it could be completely different tomorrow. And as we can see, it's obviously something that doesn't work right now. Perhaps it did at some point and then something got updated and it no longer does..?

I believe it's preferable to use a static name, something you've verified to be working. Surely, there's a chance that it will get out of date (especially if you don't take care to maintain it), but at least it's a reference to something that you know works (neither too old and broken, as it is now; nor too futuristic and broken for reasons we can't predict).

It's the same reason people pin dependencies to specific versions and not to `latest` / `master`. While today's "latest" may have been proven to work, it means nothing tomorrow.

-----------

In this line of thought, it'd be great if Docker images published to Docker hub are also versioned. Right now there's just `latest`, which is similarly dangerous.

It works for me right now, but a week into the future and it could be a completely different pgloader version (with or without backward compatibility issues) based on a completely different Debian version, etc.

df7cb:
Since bullseye is stable now, this can be closed.

Issue #1186, opened 2020-07-10
Unsigned int does not drop typemod


Sopsy:
Versions:
```
pgloader version "3.6.2"
compiled with SBCL 2.0.1.debian

mysql  Ver 8.0.20-0ubuntu0.20.04.1 for Linux on x86_64 ((Ubuntu))
psql (PostgreSQL) 12.2 (Ubuntu 12.2-4)
```

Did not try newest pgloader from master, but checked all commits after 3.6.2 and none seem to change anything related to this issue.

MySQL schema:
```mysql
CREATE TABLE `test_unsigned_serial` (
 `id` int unsigned NOT NULL AUTO_INCREMENT,
 PRIMARY KEY (`id`)
)
```

Command:
```./pgloader mysql://root:password@localhost/database postgresql://root:password@localhost/database```

Output:
```
2020-07-10T19:37:43.147000Z ERROR Database error 42601: syntax error at or near "unsigned"
QUERY: CREATE TABLE database.test_unsigned_serial
(
  id int unsigned not null
);
2020-07-10T19:37:43.147000Z FATAL Failed to create the schema, see above.
```

The following cast seems to fix it, but I'm not sure if it's a good idea or not. Obviously should fail for values larger than int32 max.
```type int when unsigned with extra auto_increment to serial drop typemod```

**Comment by Sopsy** [comment]
It actually seems even the following is enough for this.

MySQL:
```
CREATE TABLE `test_unsigned_int` (
 `id` int unsigned
)
```

Output:
```
2020-07-10T19:55:27.160000Z ERROR Database error 42601: syntax error at or near "unsigned"
QUERY: CREATE TABLE database.test_unsigned_int
(
  id int unsigned
);
2020-07-10T19:55:27.160000Z FATAL Failed to create the schema, see above.
```

Note that MySQL 8 dropped the length parameter from integers. Maybe this only occurs on MySQL 8.

**Comment by Sopsy** [comment]
Never touched any lisp code, but took a shot anyway.
This addition fixed this issue for me, but of course I have no clue what implications they might have elsewhere.

It seems MySQL no longer uses "integer", but "int" instead. Also added unsigned auto_increment int, might need to add unsigned auto_increment tinyint|smallint|mediumint|bigint, did not try those.

```
--- pgloader-bundle-3.6.2/local-projects/pgloader-3.6.2/src/sources/mysql/mysql-cast-rules.lisp 2020-03-22 21:26:18.000000000 +0000
[...]

PR #1200, opened 2020-08-27
Add MySQL unsigned int casting rules

Sopsy:
Fixes dimitri/pgloader#1186

Please comment if you think `(:source (:type "int" :unsigned t)` should be `(:source (:type "int" :unsigned t :auto-increment nil)` instead, and I can fix it.

However, the current version passed all tests I did with ints.


dimitri:
Thanks @Sopsy !

PR #1203, opened 2020-08-31
Fix ranged load from MySQL does not load the last record.

lukesilvia:
#### Problem

I tried to load MySQL tables into PostgreSQL with `rows per range` option. The following is full `with` options.

```
WITH workers = 8, concurrency = 2, multiple readers per thread, rows per range = 50000, batch rows = 10000, prefetch rows = 10000
```

But at some tables, the last record was not copied to PostgreSQL.

#### Cause

where clause for ranged load seems to ignore the last record (or the record has max primary key).

https://github.com/dimitri/pgloader/blob/9788cc64ee8014f670e0d34c853b86c82693c9ec/src/sources/mysql/mysql.lisp#L120

Thank you for checking my pull request.

### DIFF
--- a/src/sources/mysql/mysql.lisp
+++ b/src/sources/mysql/mysql.lisp
@@ -32,7 +32,7 @@
           (let* ((col (mysql-column-name
                        (nth (position coldef (table-column-list (target mysql)))
                             (fields mysql))))
-                 (sql (format nil "select min(`~a`), max(`~a`) from `~a`"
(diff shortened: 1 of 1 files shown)


dimitri:
Thanks a lot for your contribution @lukesilvia !

bendavies:
hi @dimitri,
Any chance of getting a release with this fix included?

Thanks!

PR #1209, opened 2020-09-12
Progress 4GL logical field support

wsuchy:
When creating a dump from OpenEdge (Progress) database, boolean fields will contain the following values:

- "yes" - true
- "no" - false
- "?" - null. I am assuming that everything else than yes/no is null here.

### DIFF
--- a/src/utils/transforms.lisp
+++ b/src/utils/transforms.lisp
@@ -98,8 +98,10 @@
                  varbinary-to-string
                  base64-decode
 		 hex-to-dec
+          substr
                  byte-vector-to-hexstring
-
+                 ;; progress 4gl specifics
+                 progress-logical-to-boolean 
                  ;; db3 specifics
                  logical-to-boolean
 		 db3-trim-string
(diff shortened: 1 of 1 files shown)


dimitri:
Hi, thanks for the Pull Request! Can you check if we need to return strings for this transformation? and if we must return strings, can we return "t" and "f" instead of digits?

src/utils/transforms.lisp
```diff
@@ -500,6 +503,9 @@
   "Convert a DB3 logical value to a PostgreSQL boolean."
   (if (member value '("?" " ") :test #'string=) nil value))
 
+(defun progress-logical-to-boolean (value)
+    (if (string= value "yes") "1" (if (string= value "no") "0" nil))
+)
```
dimitri, line 508: Please close all the parens on the same line. In Lisp closing parens have no semantics attached to them.

src/utils/transforms.lisp
```diff
@@ -500,6 +503,9 @@
   "Convert a DB3 logical value to a PostgreSQL boolean."
   (if (member value '("?" " ") :test #'string=) nil value))
 
+(defun progress-logical-to-boolean (value)
+    (if (string= value "yes") "1" (if (string= value "no") "0" nil))
```
dimitri, line 507: A more lispy way to write the same thing, easier to maintain and read, is the following:

``` lisp
(cond ((string-equal value "yes") "1")
      ((string-equal value "no")  "0")
      (t nil))
```

dimitri (changes requested):
Are you willing to update the PR for me to merge it?

dimitri:
Progress 4GL logical field support. The v3 codebase is in maintenance mode and this PR has outstanding change requests. If this is still relevant, please open a new issue or PR targeting the v4 codebase.

---

PR #1213, opened 2020-09-26
  Fix the header "On error stop / On error resume next"

kocio-pl:
The header "On error stop / On error resume next" contains some text that should be in the paragraph instead.


dimitri (approved):

dimitri:
Thanks @kocio-pl !

PR #1222, opened 2020-10-20
Recreate foreign keys from index. Fixes #1216

### LINKED ISSUE #1216: pgloader does not recreate all dropped foreign keys in "ORM mode"
- [X] pgloader --version
  
    ```
    3.6.2
    ```
  
  - [X] did you test a fresh compile from the source tree?
    
    Compiling pgloader from sources is documented in the
    [README](https://github.com/dimitri/pgloader#build-from-sources), it's
    easy to do, and if patches are to be made to fix your bug, you're going
    to have to build from sources to get the fix anyway…

    Yes, same result.
    
    Yes, no match

I stop here because debugging my problem lead me to what seems like a bug to me in create-pgsql-fkeys at https://github.com/dimitri/pgloader/blob/master/src/pgsql/pgsql-create-schema.lisp#L230. If my hunch is wrong, I'll try to distill my setup down to a stand alone test but that would currently take me longer than describing this.

I am using the ORM case, so `create no tables, include drop, truncate`. While pgloader drops the necessary foreign keys on the processed target tables, it fails to recreate some of them. The missing keys are mentioned as `EXTRA FK DEPS!` in the debug output but they don't get recreated.

Looking at the source of that log message, it seems to me (with my limited LISP understanding) that the second append in the top level loop in create-pgsql-fkeys does not collect any output. It has an inner loop over `index-fk-deps index` that collects the formatted sql, but that seems to be discarded in the loop over `table-index-list table`. Indeed changing the `:do` in line 246 into an `:append` solves the problem for me and recreates the missing keys.

    
  - [ ] how can I reproduce the bug?

    Incude a self-contained pgloader command file.

    If you're loading from a database, consider attaching a database dump to
    your issue. For MySQL, use `mysqldump`. For SQLite, just send over your
    source file, that's easy. Maybe be the one with your production data, of
    course, the one with just the sample of data that allows me to reproduce
    your bug.
    
    When using a proprietary database system as a source, consider creating
    a sample database on some Cloud service or somewhere you can then give
    me access to, and see my email address on my GitHub profile to send me
    the credentials. Still open a public issue for tracking and as
    documentation for other users.

```
--
-- EDIT THIS FILE TO MATCH YOUR BUG REPORT
--

LOAD CSV
     FROM INLINE with encoding 'ascii'
     INTO postgresql:///pgloader
     TARGET TABLE jordane

     WITH truncate,
          fields terminated by '|',
          fields not enclosed,
          fields escaped by backslash-quote

      SET work_mem to '128MB',
          standard_conforming_strings to 'on'

   BEFORE LOAD DO
    $$ drop table if exists jordane; $$,
    $$ CREATE TABLE jordane
       (
         "NOM" character(20),
         "PRENOM" character(20)
       )
    $$;

BORDET|Jordane
BORDET|Audrey
LASTNAME|"opening quote
BONNIER|testprenombe~aucouptroplong
JOURDAIN|héhé¶
```

  - [ ] pgloader output you obtain
  
```
PASTE HERE THE OUTPUT OF THE PGLOADER COMMAND
```

  - [ ] data that is being loaded, if relevant
  
```
PASTE HERE THE DATA THAT HAS BEEN LOADED
```

  - [ ] How the data is different from what you expected, if relevant

dfn-certling:
Fixes #1216

### DIFF
--- a/src/pgsql/pgsql-create-schema.lisp
+++ b/src/pgsql/pgsql-create-schema.lisp
@@ -243,10 +243,10 @@
                                   (fkey-foreign-table fkey))
                        :collect (format-create-sql fkey))
             :append (loop :for index :in (table-index-list table)
-                       :do (loop :for fkey :in (index-fk-deps index)
-                              :for sql := (format-create-sql fkey)
-                              :do (log-message :debug "EXTRA FK DEPS! ~a" sql)
-                              :collect sql)))))
+                       :append (loop :for fkey :in (index-fk-deps index)
+                                  :for sql := (format-create-sql fkey)
+                                  :do (log-message :debug "EXTRA FK DEPS! ~a" sql)
+                                  :collect sql)))))
     ;; and now execute our list
     (pgsql-execute-with-timing section label fk-sql-list :log-level log-level)))
 


dimitri:
FK recreation from index — the fix is appreciated but the v3 codebase is in maintenance mode. **pgloader v4 already handles FK introspection correctly** including multi-column FKs (see PR #1705). No action needed in v3.

---

PR #1238, opened 2020-11-30
Fix reading from read-only databases with recent versions of postmodern

p12tic:
Currently copying from read-only databases breaks with recent-enough version of the postmodern library.

https://github.com/marijnh/Postmodern/pull/163 changes the default transaction syntax from `BEGIN` to `BEGIN TRANSACTION ISOLATION LEVEL READ COMMITTED READ WRITE` (which corresponds to `read-committed-rw` in postmodern). This PR changes the isolation level to `read-committed-ro` for read operations.

### DIFF
--- a/src/package.lisp
+++ b/src/package.lisp
@@ -438,6 +438,7 @@
            #:pgconn-major-version
            #:pgconn-variant
            #:with-pgsql-transaction
+           #:with-pgsql-transaction-read
 	   #:with-pgsql-connection
 	   #:pgsql-execute
 	   #:pgsql-execute-with-timing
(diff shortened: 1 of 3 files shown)


mcg:
Running into this issue with a ReadOnly RDS standby instance. Any other workaround for this?

dimitri:
Fix for read-only database connections with newer PostgreSQL. This is a valid v3 fix but the codebase is in maintenance mode. **pgloader v4** sets `READ COMMITTED` isolation and handles read-only sources via standard JDBC — please verify with v4 before reopening.

---

Issue #1068, opened 2020-01-03
MSSQL to Postgres: is not of type VECTOR


wwwald:
I'm using a self-built Docker image based on pgloader version "3.6.1" (compiled with SBCL 1.4.16.debian) for reasons explained in #1057. I'm trying to convert an MSSQL database hosted on SQL Server 2017, using FreeTDS configured with protocol version 7.4 and UTF-8.

My MSSQL source database has a single "dbo" schema with about 40 tables to be converted.
By migrating the tables individually, it turns out that 5 of the 40 tables fail to convert with the following error message:
```
An unhandled error condition has been signalled:
   The value
     #.(SB-SYS:INT-SAP #X7FA0D001FE60)
   is not of type
     VECTOR
```
I've attached the debug log for one of these failed migrations here:
[pgloader.log](https://github.com/dimitri/pgloader/files/4019678/pgloader.log)

This seems similar to the problems mentioned in #896. In that case, it seems to be related to a column being marked as IDENTITY or not, but this does not explain my observations.

For now, I can't see where the table schema for these failed tables differ from other, succesfully migrating tables. It's not 1-on-1 related to them having an IDENTITY column, as #896 suggests.

Any ideas from the debug log?

Bonn93:
I got this working with TDS 8.0. See here: https://github.com/Bonn93/pgloader/blob/v3.6.1.1/Dockerfile

m0nhawk:
@Bonn93 This helped me also with a similar error. 👍

dimitri:
Closing the issue then. Please consider sending a Pull Request with a documentation update containing your findings with the TDS protocol setup.

dagostinelli:
This fixed it for me as well.  Here's what I did.

```
$ cat Dockerfile 
FROM dimitri/pgloader:latest

ADD freetds.conf /etc/freetds/freetds.conf
```
and 
```
$ cat freetds.conf 
[global]
        tds version = 8.0
        client charset = UTF-8
```
Then build it:
`$ docker build -t mypgloader .`

Then I executed it:
`$ docker run --rm --name pgloader mypgloader pgloader --dry-run --no-ssl-cert-verification mssql://<omitted> postgresql://<omitted>`

I'll make a PR for this.

dagostinelli:
https://github.com/dimitri/pgloader/pull/1301

meyerjoe:
I had the same error.

[...]

PR #1301, opened 2021-07-27
specify v8 of freetds to deal with #1068 'MSSQL to Postgres: is not o…

dagostinelli:
…f type VECTOR'


dagostinelli:
This PR is meant to enshrine the fix for the issue here: https://github.com/dimitri/pgloader/issues/1068

PR #1429, opened 2022-09-12
Added support for sequences with minvalue defined

noctarius:
When a sequence is defined with a minimum value, and the sequence to migrate is empty, the setval fails due to 1 being the default (which may be lower than the defined minimum value)


dimitri:
Thanks for your contribution @noctarius !

PR #1436, opened 2022-09-21
Handle non-standard schemas when looking up identity column

justinfalk:
The current MSSQL implementation fails to find Identity columns in non-dbo schemas.

In list-all-columns.sql the query uses
``` 
object_id(c.TABLE_NAME)
```
which limits the scope of the object query to the primary schema (dbo).  Thus, when using an alternative schema the IsIdentity column always returns null.

The fix is to include the schema when querying for object_id which will work for both primary and secondary schemas.

```
object_id(CONCAT(c.TABLE_SCHEMA, '.', c.TABLE_NAME))
```

### DIFF
--- a/src/sources/mssql/sql/list-all-columns.sql
+++ b/src/sources/mssql/sql/list-all-columns.sql
@@ -28,7 +28,7 @@
          ELSE c.COLUMN_DEFAULT
          END,
          c.IS_NULLABLE,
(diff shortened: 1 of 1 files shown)


dimitri:
Non-standard schema IDENTITY column detection. Fixed in **pgloader v4** (PR #1705) which introspects `INFORMATION_SCHEMA.COLUMNS` with full schema scope. No longer needed in v3.

---

PR #1466, opened 2023-02-17
Add mapping of getutcdate ms sql default function

### LINKED ISSUE #1343: mssql drop default or change default of timestampz columns
- [X] pgloader --version
  
    ```
    pgloader version "3.6.3~devel"
    ```
    
I have a mssql db using timestampz with a default of `getdate()` which I think translates to `getutctime()`. The default casting rules preserves this default but it is not compatible with postgres. I can't find any way to override the casting rules like what is possible with mysql.

I'm using this command file

```
load database
     from mssql://SA:{{SA_PASSWORD}}@mssql-db/mobilerecellcom
     into postgresql://postgres:postgres@db/postgres
     cast type timestamptz to timestamptz drop default

before load do $$ CREATE EXTENSION IF NOT EXISTS "uuid-ossp" $$

including only table names like 'MR_Requests', 'MR_Organization' in schema 'dbo'

alter schema 'dbo' rename to 'public'

;
```

which produces this output

`2021-12-07T15:27:50.265000Z ERROR Database error 22007: invalid input syntax for type timestamp with time zone: "getutcdate()"`

It doesn't seem like the cast rule has any affect here.

bjrne:
`getdate()` is `getutcdate()` under the hood, but `getutcdate()` can also be used and thus also needs a mapping/translation

* I compiled this locally and can confirm that this solved my use case, but this PR would probably benefit from a test. Unfortunately my lisp skills are nonexistent, so I'll leave this open for now.
* I took insipration from this PR: #577 
* This PR will probably fix #1343 which was closed without actually being fixed.

Thanks for this awesome open source software!


dimitri:
Mapping `GETUTCDATE()` MSSQL default to PostgreSQL. The v3 codebase is in maintenance mode. Column defaults from MSSQL are handled in **pgloader v4** — if this specific function is not mapped, please open a new issue against v4.

---

PR #1472, opened 2023-03-15
SQLite: check for base64 for text content in blob

### LINKED ISSUE #415: ERROR Bad character - in base64 decode
This seems related to #60 

I'm using a CI/CD tool called Drone and started using it with sqlite3.  I'm trying to migrate to postgresql.

When using pgloader to migrate the data I see:

```
2016-05-26T02:57:20.711000Z ERROR Bad character - in base64 decode
```

It seems when it tries to migrate data in sqlite3 table like: https://github.com/drone/drone/blob/v0.4.2/store/datastore/ddl/sqlite3/1_init.sql#L50-L57 

That it fails to properly decode the data.  As you can see the BLOBs are public/private keys that contain '-' in them.

The postgresql schema is defined like: https://github.com/drone/drone/blob/v0.4.2/store/datastore/ddl/postgres/1_init.sql#L40-L47

Just curious if you had any thoughts on how to workaround this issue.

Thanks!

darioleanbit:
If the SQLite driver returns the value of a blob column a string, don't make the assumption that the value is base64-encoded. Instead, check for actual base64 content by trying to decode the string as base64.

Will probably fix #415, which is closed, but not really solved.

### DIFF
--- a/src/sources/sqlite/sqlite.lisp
+++ b/src/sources/sqlite/sqlite.lisp
@@ -4,6 +4,17 @@
 
 (in-package :pgloader.source.sqlite)
 
+(in-package :pgloader.source.sqlite)
+
+(declaim (inline has-base64-sequence-p))
+
+;;; Return the decoded value of a base64 string, ignoring errors.
+;;; Returns nil if the string doesn't contain a valid base64 string.
+;;;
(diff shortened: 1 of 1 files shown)


dimitri:

```
```

Closing as fixed in v4.

PR #1482, opened 2023-05-01
add IDE specific gitignores.

aadrian:
add IDE specific gitignores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,8 @@ web/howto/sqlite.html
 system-index.txt
 buildapp.log
 docs/_build
+
+# IDE specific ignores
+.vscode/
+.idea/
+.settings/


dimitri:
Thanks for the PR. We prefer not to add IDE-specific patterns to the project's `.gitignore` — that file is for build artifacts and generated files that every contributor needs to ignore regardless of editor.

Git has two built-in mechanisms for personal ignores that do not touch the shared file:

**Per-repository** (stored in `.git/info/exclude`, never committed):
```
echo '.idea/' >> .git/info/exclude
```

**Global** (applies to every repository on your machine — the right place for IDE patterns):
```
git config --global core.excludesFile ~/.gitignore_global
echo '.idea/'     >> ~/.gitignore_global
echo '.vscode/'   >> ~/.gitignore_global
echo '.settings/' >> ~/.gitignore_global
```

Closing in favour of each contributor setting this up locally.

Issue #425, opened 2016-06-20
resetting sequences fails with quoted identifiers


leoschweizer:
I run the  following command to migrate my database from MySQL to Postgres:
```
$ docker run --rm --name pgloader dimitri/pgloader:latest pgloader --verbose --with 'quote identifiers' mysql://.../old postgresql://.../new
```
Everything works fine if I omit the 'quote identifiers' option, however I have to preserve the casing of my column names. Running with this options yields the following error:

```
2016-06-20T13:23:30.644000Z NOTICE Reset sequences
An unhandled error condition has been signalled:
   Database error 42703: column "taskid" does not exist
HINT: Perhaps you meant to reference the column "Task.taskId".
CONTEXT: PL/pgSQL function inline_code_block line 24 at EXECUTE
QUERY:
DO $$
DECLARE
  n integer := 0;
  r record;
BEGIN
  FOR r in
       SELECT 'select '
               || trim(trailing ')'
                  from replace(pg_get_expr(d.adbin, d.adrelid),
                               'nextval', 'setval'))
               || ', (select greatest(max(' || a.attname || '), 1) from only '
               || quote_ident(nspname) || '.' || quote_ident(relname) || '));' as sql
         FROM pg_class c
              JOIN pg_namespace n on n.oid = c.relnamespace
              JOIN pg_attribute a on a.attrelid = c.oid
              JOIN pg_attrdef d on d.adrelid = a.attrelid
                                 and d.adnum = a.attnum
                                 and a.atthasdef
        WHERE relkind = 'r' and a.attnum > 0
              and pg_get_expr(d.adbin, d.adrelid) ~ '^nextval'
              and c.oid in (select oid from reloids)
  LOOP
    n := n + 1;
    EXECUTE r.sql;
  END LOOP;

  PERFORM pg_notify('seqs', n::text);
END;
$$;                                                                ^
```
To me, it seems like `a.attname` is not quoted properly in this query. In the Postgres output I can see that it tries to read `taskId`, without quotes:
```
QUERY:  select setval('"Task_taskId_seq"'::regclass, (select greatest(max(taskId), 1) from only p
ublic."Task"));
```

This is my exact pgloader version:
```
$ docker run --rm --name pgloader dimitri/pgloader:latest pgloader --version
WARNING:
Couldn't re-execute SBCL with proper personality flags (/proc isn't mounted? setuid?)
Trying to continue anyway.
pgloader version "3.3.0.51~devel"
[...]

PR #1509, opened 2023-07-16
Fix "reset sequences" crash on tables with case sensitive column names

codereverser:
On case sensitive migrations (i.e. with "quote identifiers"), reset sequences fail with the following error. 

```
Database error 42703: column ""<columnName>"" of relation "<table>" does not exist
```

This is due to invalid parameters being passed to `pg_get_serial_sequence` function introduced in  e32066d35a03df3b386189d5c4e9dcd11805ee03

for example, on a mysql table with following definition

```SQL
CREATE TABLE botLog(
    `logID` bigint(20) NOT NULL AUTO_INCREMENT, 
    `message` varchar(255), 
    PRIMARY KEY (`logID`)
);
```

```SQL
SELECT pg_get_serial_sequence(quote_ident('public') || '.' || quote_ident('botLog'), quote_ident('logID'));
ERROR:  column ""logID"" of relation "botLog" does not exist
```
throws the same error, but the following works

```SQL
SELECT pg_get_serial_sequence(quote_ident('public') || '.' || quote_ident('botLog'), 'logID');
  pg_get_serial_sequence
---------------------------
 public."botLog_logID_seq"
 ```

ref: https://github.com/dimitri/pgloader/issues/425

### DIFF
--- a/src/pgsql/pgsql-create-schema.lisp
+++ b/src/pgsql/pgsql-create-schema.lisp
@@ -409,7 +409,7 @@ BEGIN
                   from replace(pg_get_expr(d.adbin, d.adrelid),
                                'nextval', 'setval'))
                || ', (select greatest(max(' || quote_ident(a.attname) || '), (select seqmin from pg_sequence where seqrelid = ('''
-               || pg_get_serial_sequence(quote_ident(nspname) || '.' || quote_ident(relname), quote_ident(a.attname)) || ''')::regclass limit 1), 1) from only '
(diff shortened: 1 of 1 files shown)


Asher-JH:
Is this something that can be merged in soon? facing this issue

dimitri:
Closing — this fix landed in main as part of #1717 (commit `7938678`). The `pg_get_serial_sequence` call now passes `a.attname` directly without `quote_ident()`, which is exactly what this PR proposed. Thanks for tracking it down!

PR #1514, opened 2023-07-28
Mysql datetime has no time zone.

sdyarnell:
To ensure consistent behavior across applications, database types should remain as consistent as possible. 

MySQL datetime does not have a time zone. The appropriate PostgreSQL type is timestamp.

Caveat: I have no clue what I'm doing.

### DIFF
--- a/src/sources/mysql/mysql-cast-rules.lisp
+++ b/src/sources/mysql/mysql-cast-rules.lisp
@@ -125,19 +125,19 @@
     ;; MySQL and dates...
     ;;
     (:source (:type "datetime" :default "0000-00-00 00:00:00" :not-null t)
-     :target (:type "timestamptz" :drop-default t :drop-not-null t)
+     :target (:type "timestamp" :drop-default t :drop-not-null t)
      :using pgloader.transforms::zero-dates-to-null)
 
     (:source (:type "datetime" :default "0000-00-00 00:00:00")
-     :target (:type "timestamptz" :drop-default t)
+     :target (:type "timestamp" :drop-default t)
      :using pgloader.transforms::zero-dates-to-null)
 
     (:source (:type "datetime" :on-update-current-timestamp t :not-null t)
-     :target (:type "timestamptz" :drop-default t :drop-not-null t)
+     :target (:type "timestamp" :drop-default t :drop-not-null t)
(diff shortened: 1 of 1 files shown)


df7cb:
-1. The default aka sane timestamp type to use in PostgreSQL is timestamptz, and if Mysql lacks the TZ info, migrating to PG is the perfect opportunity to add it. (Note that timestamptz doesn't actually store the timezone, it just adds it for input/output.)

https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_timestamp_.28without_time_zone.29

syarnellen:
The issue would be going throughout the massive application to account for this change in the database. 

As this is a database migration tool, database optimizations based on postgres recommendations do not feel like they align with the tool when they alter the behavior.

Also, I could not get pgloader to parse the following, or I probably would have never got here lol --

type datetime when default "0000-00-00 00:00:00" and not null
  to timestamp drop not null drop default
      using zero-dates-to-null

Is it because there are 2 guards?

sdyarnell:
The issue would be going throughout the massive application to account for this change in the database.

As this is a database migration tool, database optimizations based on postgres recommendations do not feel like they align with the tool when they alter the behavior.

Also, I could not get pgloader to parse the following, or I probably would have never got here lol --

type datetime when default "0000-00-00 00:00:00" and not null
to timestamp drop not null drop default
using zero-dates-to-null

Is it because there are 2 guards?

dimitri:
Thank you for taking the time to investigate this and open a PR — the semantic argument is actually correct: MySQL `datetime` stores no timezone information, and PostgreSQL `timestamp without time zone` is the faithful counterpart, while MySQL `timestamp` (UTC-stored, session-tz on read) more closely mirrors `timestamptz`.

That said, we're not going to make this change as a default. The current mapping of `datetime` → `timestamptz` is intentional: it reflects the way most MySQL applications actually use the type (UTC values regardless of MySQL's lack of TZ storage), and `timestamptz` is the safer default for migrated schemas. Changing it would silently break existing migrations.

If you need the more faithful mapping, it's already achievable today with a cast rule in your load file:

```
CAST type datetime to timestamp drop typemod
```

Thanks again for the contribution.

Issue #1327, opened 2021-10-13
`preserve index names` doesn't respect `quote identifiers`


Faithfinder:
- [x] pgloader --version
  
    ```
    3.6.1
    ```
  
  - [ ] did you test a fresh compile from the source tree?
    
  - [x] did you search for other similar issues?
    
  - [x] how can I reproduce the bug?

Table definition:

```sql
CREATE TABLE `test` (
  `Indexed` int(11) DEFAULT NULL,
  UNIQUE KEY `test_Indexed_IDX` (`Indexed`) USING BTREE
)
```

command.load
```
LOAD DATABASE
     FROM      mysql://root:root@localhost/test
     INTO      postgresql://root:root@localhost:5432/test

WITH quote identifiers, preserve index names
```

`test_Indexed_IDX` -> `test_indexed_idx`

amvandenberg:
Ran into the same issue.

nikolaigut:
I also have the same problem.

PR #1521, opened 2023-08-25
preserve upper and lower case of an index

nikolaigut:
this change preserves the upper and lower case of an index to be migrated

https://github.com/dimitri/pgloader/issues/1327

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,9 @@
 FROM debian:stable-slim as builder
 
+  RUN apt-get update && apt-get install -y wget
+  RUN wget http://archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1-1ubuntu2.1~18.04.23_amd64.deb
+  RUN dpkg -i libssl1.1_1.1.1-1ubuntu2.1~18.04.23_amd64.deb
+
   RUN apt-get update \
       && apt-get install -y --no-install-recommends \
         bzip2 \
@@ -9,7 +13,6 @@ FROM debian:stable-slim as builder
         gawk \
         git \
         libsqlite3-dev \
-        libssl1.1 \
         libzip-dev \
         make \
         openssl \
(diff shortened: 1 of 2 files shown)


dimitri:
Thank you for tracking this down and submitting a fix — case-sensitive index names are a real pain point in MySQL→PostgreSQL migrations and this is a clean, targeted diagnosis.

We've taken the fix forward in [#1729](https://github.com/dimitri/pgloader/pull/1729) with one small adaptation: instead of embedding literal double-quote characters in the format string, we used the `~s` format directive (Common Lisp's `prin1`, which wraps a string in double quotes). This is consistent with how the rest of the codebase quotes identifiers in SQL — see `CREATE EXTENSION IF NOT EXISTS ~s` and `ALTER DATABASE ~s` in `pgsql-create-schema.lisp`.

The Dockerfile change from your PR was not carried over — `libssl1.1` has since been replaced by `libssl3` directly in the apt-get install list.

Closing this in favour of #1729. Thanks again!

dimitri:
Thank you for the fix and for digging up the workaround from the issue thread — this is a real pain point for anyone migrating MySQL schemas with ENUM/SET columns.

We took a slightly different approach in [#1730](https://github.com/dimitri/pgloader/pull/1730):

- The double-underscore separator (`table__column`) was not chosen because it still allows collisions: a table named `foo__bar` with an ENUM column `baz` would produce type `foo__bar__baz`, which can still clash with a table named `foo__bar__baz`. The collision space is smaller, but the problem is not fundamentally fixed.
- Instead, we append a `_t` suffix to make the type name `{table}_{column}_t`. Since table names virtually never end in `_t`, this eliminates the common case cleanly.
- We also added a runtime check against the target PostgreSQL `pg_type` catalog before issuing `CREATE TYPE`, so that if `_t` somehow still conflicts, pgloader falls back to `_enum` suffix and then `enum_` prefix — and errors out clearly if all three candidates are taken.
- The fix covers both v3 and v4.

Thanks again for reporting and for the patch!

Issue #1486, opened 2023-05-08
quote identifiers not applied to foreign key constraints from sqlite3


jinnatar:
- [x] pgloader --version
    ```shell
    %> pgloader-bundle-3.6.9/bin/pgloader --version
    pgloader version "3.6.9"
    compiled with SBCL 2.1.1.debian
    ```
  
  - [x] did you test a fresh compile from the source tree?
    ```shell
    %> build/bin/pgloader --version
    pgloader version "3.6.999791d"
    compiled with SBCL 2.1.1.debian
    ```
  - [x] did you search for other similar issues?
    
  - [x] how can I reproduce the bug?
    Dump from sqlite db.sqlite3:
```
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE pages (id int PRIMARY KEY, authorId int, FOREIGN KEY(authorId) REFERENCES users("authorId"));
INSERT INTO pages VALUES(1,42);
CREATE TABLE users (id bigint PRIMARY KEY, name text NOT NULL, authorId int NOT NULL);
INSERT INTO users VALUES(1,'demo',42);
COMMIT;
```
 

  - [x] pgloader output you obtain
  
```
%> PGPASSWORD=foobar build/bin/pgloader --with "quote identifiers" sqlite://db.sqlite3 pgsql://postgres@postgres.service.consul/demo
2023-05-08T11:50:21.016000+03:00 LOG pgloader version "3.6.999791d"
2023-05-08T11:50:21.080002+03:00 LOG Migrating from #<SQLITE-CONNECTION sqlite:///vol/home/src/3rd-party-forks/pgloader/db.sqlite3 {10074EE4C3}>
2023-05-08T11:50:21.080002+03:00 LOG Migrating into #<PGSQL-CONNECTION pgsql://postgres@postgres.service.consul/demo {1007644B63}>
2023-05-08T11:50:21.600015+03:00 ERROR PostgreSQL Database error 42703: column "authorid" referenced in foreign key constraint does not exist
QUERY: ALTER TABLE "pages" ADD FOREIGN KEY(authorId) REFERENCES "users"(authorId) ON UPDATE NO ACTION ON DELETE NO ACTION
2023-05-08T11:50:21.608015+03:00 LOG report summary reset
             table name     errors       rows      bytes      total time
-----------------------  ---------  ---------  ---------  --------------
                  fetch          0          0                     0.000s
        fetch meta data          0          5                     0.064s
[...]

PR #1531, opened 2023-09-18
fix(sqlite): respect case option for foreign keys

Redlinkk:
When using the option "quote identifiers", I had errors ("column doesn't exist") when importing foreign keys constraints from sqlite to PG. This should fix the issue

### DIFF
--- a/src/sources/sqlite/sqlite-schema.lisp
+++ b/src/sources/sqlite/sqlite-schema.lisp
@@ -241,8 +241,8 @@
                                   pg-fkey)))))
              (if (and fkey from to)
                  (progn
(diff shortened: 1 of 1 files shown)


ralgar (approved):
Can confirm that this works as expected. Solves my issue from #1486.

**EDIT:** If anyone wants a Docker image built from this PR, you can use `ghcr.io/ralgar/pgloader:pr-1531`

hoelter:
I can also confirm this works as expected. I was running into the mentioned error when using "with quote identifiers" on a sqlite database, and this change fixed that error with constraints migrating as expected.

dimitri:

```
```

Closing as fixed in v4.

Issue #1551, opened 2023-12-20
MSSQL Materialize view not working


harshagarwal001:
Thanks for contributing to [pgloader](https://pgloader.io) by reporting an
issue! Reporting an issue is the only way we can solve problems, fix bugs,
and improve both the software and its user experience in general.

The best bug reports follow those 3 simple steps:

  1. show what you did,
  2. show the result you got,
  3. explain how the result is not what you expected.

In the case of pgloader, here's the information I will need to read in your
bug report. Having all of this is a big help, and often means the bug you
reported can be fixed very efficiently as soon as I get to it.

Please provide the following information:

  - [ ] pgloader --version
  
    ```
    <fill pgloader version here>
    ```
  
  - [ ] did you test a fresh compile from the source tree?
    
    Compiling pgloader from sources is documented in the
    [README](https://github.com/dimitri/pgloader#build-from-sources), it's
    easy to do, and if patches are to be made to fix your bug, you're going
    to have to build from sources to get the fix anyway…
    
  - [ ] did you search for other similar issues?
    
  - [ ] how can I reproduce the bug?

    Incude a self-contained pgloader command file.

    If you're loading from a database, consider attaching a database dump to
    your issue. For MySQL, use `mysqldump`. For SQLite, just send over your
    source file, that's easy. Maybe be the one with your production data, of
    course, the one with just the sample of data that allows me to reproduce
    your bug.
    
    When using a proprietary database system as a source, consider creating
    a sample database on some Cloud service or somewhere you can then give
    me access to, and see my email address on my GitHub profile to send me
    the credentials. Still open a public issue for tracking and as
    documentation for other users.

```
--
-- EDIT THIS FILE TO MATCH YOUR BUG REPORT
--

load database
     from mssql://USER_NAME@@3$@SOME_IP/db
     into postgresql:///db

CAST type nvarchar to text drop typemod using remove-null-characters,
      type varchar to text drop typemod using remove-null-characters

MATERIALIZE VIEWS view1
including only table names like 'tb1', 'tb2' in schema 'dbo'

set work_mem to '4GB', maintenance_work_mem to '4GB'

before load do $$ drop schema if exists dbo cascade; $$,
 $$ CREATE EXTENSION IF NOT EXISTS "uuid-ossp" $$

ALTER schema 'dbo' rename to 'public';

BORDET|Jordane
BORDET|Audrey
LASTNAME|"opening quote
BONNIER|testprenombe~aucouptroplong
[...]

PR #1559, opened 2024-02-21
fix: mssql should only load data for the views specified by materialize views

conoremclaughlin:


svantevonerichsen6906 (changes requested):
Please don't litter the code base with debugging prints, and please do adhere to the formatting standard that is very much established in the entire Lisp language ecosystem.  Can't review like this.

conoremclaughlin:
Hi @svantevonerichsen6906, thanks for the comments. I wanted to get the PR up for anyone who found it useful before cleaning it up.

I've removed the debug statements and have changed the parentheses to follow more of Lisp's standards. Unfortunately it's been quite some time since I learned Lisp in Berkeley and Lisp doesn't have a good formatter, so this is about as far as I can go. Otherwise I would run the formatter across the file. The original code specifically notes that this flow was untested by the author, so anything is an improvement over the untested / completely broken materialize views code path we currently have for MSSQL.

Thanks!

conoremclaughlin:
Fixes #1551.

conoremclaughlin:
Bumping this @svantevonerichsen6906. Thanks man!

src/sources/mssql/mssql.lisp
```diff
@@ -69,17 +69,28 @@
...
+        (loop :for (schema-name . view-name) :in view-names
+          :do (let* (
+              (schema-name (or schema-name "dbo"))
+              (schema-entry (or
+                (assoc schema-name including :test #'string=)
+                (let
+                  ((new-entry (cons schema-name nil))); Initially nil, intending to be a list
+                  (push new-entry including)
+                  new-entry))))
+          (push-to-end view-name (cdr schema-entry))))
```
svantevonerichsen6906, line 93: What I mean is: completely clean up the debugging alterations, and this is the indentation that should appear here:

```
(let* ((view-names (unless (eq :all materialize-views)
                     (mapcar #'matview-source-name materialize-views)))
       (including nil))
  (loop :for (schema-name . view-name) :in view-names
        :do (let* ((schema-name (or schema-name "dbo"))
                   (schema-entry (or
                                  (assoc schema-name including :test #'string=)
                                  (let ((new-entry (cons schema-name nil)))
                                    (push new-entry including)
                                    new-entry))))
              (push-to-end view-name (cdr schema-entry))))
  …)
```

I don't fully understand what's changed here yet, though

dimitri:
Thanks for the fix, @conoremclaughlin — the diagnosis was correct.

This was properly resolved as part of a broader cleanup in #1720 (merged 2026-06-23). The clean fix uses a separate `view-including` alist built from `(copy-tree including)` plus the view names, leaving the outer `including` filter untouched for `fetch-indexes` and `fetch-foreign-keys`. A regression test (`filtertest` database, `09-filter-views-counts`, `10-filter-views-integrity`) was added at the same time to prevent it regressing.

Closing in favour of that work.

Issue #1245, opened 2020-12-17
pgloader create types with same name than tables


4ng31:
Hello, 

I was trying to migrate a MySQL database into PostgresSQL, and I found this error between generated data type and tables name.

Example:

In the origin database there is a table called "acreditacion_haberes" with a corresponding type called "estado" , pgloader creates a TYPE called "acreditacion_haberes_estado". The problem occurs because in the origin database exist another  tables called  "acreditacion_haberes_estado", wich is the same name. Also this other table has a corresponding TYPE called "estado", and pgloder creates another TYPE called "acreditacion_haberes_estado_estado"

 At the end there is a TYPE and a TABLE with the same names, and this apperently is not allowed in pg_class.

2020-12-17T22:25:07.996000+06:00 SQL DROP TYPE IF EXISTS gem.acreditacion_haberes_estado CASCADE;
2020-12-17T22:25:07.997000+06:00 SQL CREATE TYPE gem.acreditacion_haberes_estado AS ENUM ('Pendiente', 'Rechazado', 'Autorizado', 'Existe');

2020-12-17T22:25:08.738000+06:00 ERROR Database error 42710: type "acreditacion_haberes_estado" already exists
HINT: A relation has an associated type of the same name, so you must use a name that doesn't conflict with any existing type.
QUERY: CREATE TABLE gem.acreditacion_haberes_estado 
(
  id                      bigserial not null,
  acreditacion_haberes_id bigint not null,
  estado                  gem.acreditacion_haberes_estado_estado,
  fecha                   date,
  usuario_id              bigint not null,
  motivo                  varchar(200),
  audi_user               bigint,
  audi_fecha              timestamptz,
  audi_accion             gem.acreditacion_haberes_estado_audi_accion
);
2020-12-17T22:25:08.739000+06:00 FATAL Failed to create the schema, see above.

  - [ ] pgloader --version
  
    ```
pgloader version "3.6.2"
compiled with SBCL 2.0.1-4.el8
    ```
  
  - [*] did you test a fresh compile from the source tree?

   
  - [ *] did you search for other similar issues?
    
  - [ ] how can I reproduce the bug?

    Incude a self-contained pgloader command file.

4ng31:
I've just recompile pgloader changing the MySQL data type casting rules, adding another underscore (_)

src/sources/mysql/mysql-cast-rules.lisp 
Line 10:
(format nil "~a_~a" (unquote table-name #\") (unquote column-name #\"))))
[...]

PR #1567, opened 2024-03-19
Fix: pgloader creates table with the same name as type.

Chirikumbrah:
Took code from this [reply](https://github.com/dimitri/pgloader/issues/1245#issuecomment-748763652) :)

### DIFF
--- a/src/sources/mysql/mysql-cast-rules.lisp
+++ b/src/sources/mysql/mysql-cast-rules.lisp
@@ -7,7 +7,7 @@
 (defun enum-or-set-name (table-name column-name type ctype typemod)
   (declare (ignore type ctype typemod))
   (apply-identifier-case
-   (format nil "~a_~a" (unquote table-name #\") (unquote column-name #\"))))
+   (format nil "~a__~a" (unquote table-name #\") (unquote column-name #\"))))
 
 ;;;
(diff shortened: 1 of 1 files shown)


dimitri:
Thank you for the fix and for digging up the workaround from the issue thread — this is a real pain point for anyone migrating MySQL schemas with ENUM/SET columns.

We took a slightly different approach in [#1730](https://github.com/dimitri/pgloader/pull/1730):

- The double-underscore separator (`table__column`) was not chosen because it still allows collisions: a table named `foo__bar` with an ENUM column `baz` would produce type `foo__bar__baz`, which can still clash with a table named `foo__bar__baz`. The collision space is smaller, but the problem is not fundamentally fixed.
- Instead, we append a `_t` suffix to make the type name `{table}_{column}_t`. Since table names virtually never end in `_t`, this eliminates the common case cleanly.
- We also added a runtime check against the target PostgreSQL `pg_type` catalog before issuing `CREATE TYPE`, so that if `_t` somehow still conflicts, pgloader falls back to `_enum` suffix and then `enum_` prefix — and errors out clearly if all three candidates are taken.
- The fix covers both v3 and v4.

Thanks again for reporting and for the patch!

PR #1608, opened 2024-09-01
fix (mssql): quote indexes whose names contain e.g. "-"

mayerrobert:
When moving data MSSQL -> Postgres indexes with names containing a dash gave errors.

### DIFF
--- a/src/sources/mssql/mssql-schema.lisp
+++ b/src/sources/mssql/mssql-schema.lisp
@@ -98,7 +98,8 @@
                             incl-where
                             excl-where  ; do we print the clause?
                             excl-where))
-     :do (let* ((schema     (find-schema catalog schema-name))
+     :do (let* ((index-name (apply-identifier-case index-name))
+                (schema     (find-schema catalog schema-name))
                 (table      (find-table schema table-name))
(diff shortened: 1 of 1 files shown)


dimitri:
Thanks for tracking this down, @mayerrobert. The reported error (dash in MSSQL index name causing a PostgreSQL DDL failure) is not reproducible on current code.

I added a regression test in [#1736](https://github.com/dimitri/pgloader/pull/1736) to verify this. Both migration paths handle dashes correctly:

- **Default (uniquify):** `build-identifier` constructs `idx_{oid}_idx-dash-test`, then `apply-identifier-case` detects the dash, forces `:quote`, and emits `CREATE INDEX "idx_19843_idx-dash-test"`
- **PRESERVE INDEX NAMES:** `apply-identifier-case` is called directly on the raw name and similarly produces `CREATE INDEX "idx-dash-test"`

Both produce valid, double-quoted PostgreSQL identifiers — 0 errors.

The fix proposed here (applying `apply-identifier-case` at schema-load time in `mssql-schema.lisp`) would be redundant since the DDL phase already covers it. Closing in favour of the test-only PR.

PR #1609, opened 2024-09-01
also omit "DEFAULT NULL" when (eq (column-default column) :NULL)

mayerrobert:
pgloader omits "DEFAULT NULL" clauses when (column-default column) is NIL. With this patch "DEFAULT NULL" clauses are also omitted when (column-default column) is :NULL.

### DIFF
--- a/src/pgsql/pgsql-ddl.lisp
+++ b/src/pgsql/pgsql-ddl.lisp
@@ -137,7 +137,7 @@
           (column-type-mod column)
           (column-type-mod column)
           (column-nullable column)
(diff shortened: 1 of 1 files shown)


dimitri:
Thanks for this fix, @mayerrobert — the diagnosis is correct.

I rebased your commit onto current main (which had a conflict with the `effective-type-mod` refactor that landed after you submitted) and opened [#1737](https://github.com/dimitri/pgloader/pull/1737) with attribution preserved. Closing this one in favour of that PR.

PR #1613, opened 2024-09-18
Use latest cosign action to fix signing issue on docker publish

arajkumar:
Adopts changes from pgcopydb https://github.com/dimitri/pgcopydb/blob/main/.github/workflows/docker-publish.yml


arajkumar:
@dimitri Here is a reference to the passing workflow with this fix => https://github.com/arajkumar/pgloader/actions/runs/10921708057/job/30314422180

dimitri (approved):

PR #1614, opened 2024-10-01
Remove rows per range option from Postgres-Postgres docs

### LINKED ISSUE #1550: Option "rows per range" not working in PostgreSQL to PostgreSQL migration?
Hello! 
I need your help. Documentation for PostgreSQL to PostgreSQL migration claims that I can use "rows per range" option:
```
* rows per range
How many rows are fetched per SELECT query when using multiple readers per thread, see above for details.
```
But pgloader of multiple versions (see below) refuses to parse config file if I add these options - I always get `ESRAP-PARSE-ERROR`
for both `multiple readers per thread` and `rows per range` commands. Without them config file works.

I have huge migration pg2pg where I would really need to use these options, but pgloader refuses to parse them. Here is WITH part of my config file:
```
    WITH
        include drop,
        create tables,
        workers = 8,
        concurrency = 4, 
        multiple readers per thread,
        rows per range 1000000
```
Am I doing something wrongly? Is there some workaround? 
Thank you very much for your help!

***

  - [X] pgloader --version
```
* from debian package - pgloader (3.6.10-1.pgdg20.04+1) - shows me version:
pgloader version "3.6.7~devel"
compiled with SBCL 2.1.1.debian

&
* manual build from latest release sources tar.gz on testing env:
pgloader version "3.6.9"
compiled with SBCL 2.0.1.debian

&
* manual build from latest master - which surprisingly shows me version:
pgloader version "3.6.7~devel"
compiled with SBCL 2.0.1.debian
(I tried to download and build repeatedly - master really shows me this version)
```
  
  - [X] did you test a fresh compile from the source tree?
    Yes, I did
    
    Yes I did but did not find anything relevant

  - [X] how can I reproduce the bug?
    Just try those options in pg2pg migration

  - [X] pgloader output you obtain
  
```
KABOOM!
ESRAP-PARSE-ERROR: At

          workers = 8,
          concurrency = 4, 
  
                         ^ (Line 9, Column 23, Position 187)

In context PGSQL-OPTIONS:

While parsing PGSQL-OPTIONS. Expected:

     the character Tab
  or the character Newline
  or the character Return
  or the character Space
  or the string "--"
  or the string "/*"
  or the string "batch"
  or the string "concurrency"
  or the string "create"
  or the string "data"
  or the string "disable"
  or the string "downcase"
  or the string "drop"
  or the string "foreign"
  or the string "include"
  or the string "max"
  or the string "no"
  or the string "on"
  or the string "prefetch"
  or the string "preserve"
  or the string "quote"
  or the string "reset"
  or the string "schema"
  or the string "snake_case"
  or the string "truncate"
  or the string "uniquify"
  or the string "workers"
An unhandled error condition has been signalled: At
```

  - [X] data that is being loaded, if relevant
  not relevant, pgloader refuses to parse config file

Thank you very much for your help!

mbanck-cd:
This was never implemented for pgsql<->pgsql migrations.

Fixes: #1550

### DIFF
--- a/docs/ref/pgsql.rst
+++ b/docs/ref/pgsql.rst
@@ -212,11 +212,6 @@ supported, and the default *WITH* clause is: *no truncate*, *create schema*,
     When this option is listed pgloader only issues the `COPY` statements,
     without doing any other processing.
 
-  - *rows per range*
-  
-    How many rows are fetched per `SELECT` query when using *multiple
-    readers per thread*, see above for details.
-
 PostgreSQL Database Casting Rules
 ---------------------------------
 


dimitri:
Closing — the stub documentation this PR removes has since been replaced in `main` with accurate content: `rows per range` is now documented as accepted for compatibility but superseded by the `chunk size` option. The intent of this PR is already addressed. Thanks for catching the original inaccuracy!

PR #1626, opened 2024-11-16
fix (typo): Fix typo in install.rst

farajael:


dimitri:
Thanks for catching this, @farajael — the fix has since been applied to `master` via a different path, so the typo is already gone. Closing as superseded.

PR #1644, opened 2025-02-14
Make docker stop emitting FromAsCasing warning

dilbernd:
It’s really nothing, just preventing an annoying warning.


makeitokay (approved):
LGTM

dimitri:
Closing — the `AS` capitalisation fix is already present in `main` (the Dockerfile already has `FROM debian:bookworm-slim AS builder`). Thanks for the contribution!

PR #1662, opened 2025-04-11
fix: remove batch concurrency support

### LINKED ISSUE #1661: prefetch rows parameter is overriden by batch concurrency (which is deprecated)
My options : 
```
with include drop, truncate, create tables, create indexes, batch rows = 10000, batch concurrency = 3, prefetch rows = 20000,
  reset sequences,
  downcase identifiers
```

Ending with a prefetch rows of 3 for reader buffer.

we should drop this option support

Sarcouy:
Fixes https://github.com/dimitri/pgloader/issues/1661

if `batch concurrency` option is specified, it'll override `prefetch rows`, since `batch concurrency` is deprecated, I dropped it.

### DIFF
--- a/src/parsers/command-options.lisp
+++ b/src/parsers/command-options.lisp
@@ -73,13 +73,10 @@
     (bind (((_ _ _ val) batch-size))
       (cons :batch-size val))))
 
-;;; deprecated, but still accept it in the parsing
-(defrule option-prefetch-rows (and (or (and kw-batch kw-concurrency)
-                                       (and kw-prefetch kw-rows))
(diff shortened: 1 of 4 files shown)


dimitri:
Superseded by #1742, which takes a more conservative approach: instead of removing the `batch concurrency` alias entirely (breaking load files that use it alone), #1742 keeps it as a deprecated alias but errors at parse time when both `batch concurrency` and `prefetch rows` appear in the same `WITH` clause. This gives a clear error message for the conflict described in #1661 without being a breaking change for users of the deprecated form.

PR #1674, opened 2025-06-03
Eliminate docker build casing warning

fossum:
There is a small warning when building the Docker image. This small change fixes the warning.


fossum:
It'd be nice to see the latest version in DockerHub

Issue #1673, opened 2025-05-27
downcase identifiers


samaneh-kamalian:
Hi. thanks for your useful package.
I want to move my database of MSSQL to postgresql. I use pgloader ms.load command
every thing is ok. but pgloader change all my database name to lowercase and I do not want this .
I search and say use  quote identifiers to prevent lowercase. but this option add addition double quotes to my name and I get error . for example : 

UNDEFINED-COLUMN: Database error 42703: column ""ID"" of relation "TNewPolicyMainTable" does not exist

 how can  disable convert to lowercase without adding double quotes?

I realy stpped on it . thanks in advance for you attention.

samaneh-kamalian:
Is anyone there? :(

samaneh-kamalian:
Please answer me how to remove this default behavior of pgloader to prevent downcase identifiers. :(

samaneh-kamalian:
it fixed in this way. I use quote identifiers option and change src/pgsql/pgsql-create-schema.lisp function reset-sequences. 
in this way: 
in this funciton there is two quote_ident(a.attname) , I change second one to remove quote_ident and just be a.attname. 
and this fix my bug.

pgloader version "3.6.af8c3c1"

til:
Thanks for suggesting this fix, it works for me. Here is a PR with it: https://github.com/dimitri/pgloader/pull/1701

dimitri:
The root cause: `pg_get_serial_sequence(table, column)` expects a **plain column name** as its second argument, not a SQL identifier expression. The `reset-sequences` DO block was passing `quote_ident(a.attname)` there — so for a column named `Id` it passed `"Id"` (with literal double-quote characters) — causing the function to look for a non-existent column named `"Id"` (quotes included as part of the name). The broken sequence name then propagated back into the dynamic SQL, producing the `""ID""` double-quoting seen in the error.

The `quote_ident` is correct in the `max(...)` expression (which IS a SQL expression context), but wrong as the column-name argument to `pg_get_serial_sequence`.

This was introduced by PR #1429 and fixed by PR #1706 — the `pg_get_serial_sequence` call now passes `a.attname` directly. That fix is on `master` now.

v4 handles this correctly by design: `reset-sequences-sql` uses the raw `col-name` for `pg_get_serial_sequence` and `identifier-quote col-name` only for the `MAX(...)` expression, keeping the two contexts cleanly separated.

Issue #1651, opened 2025-03-19
Column names getting double-quoted when importing from sqlite using --with "quote identifiers"


funkypenguin:
Hi,

Beginner here - I'm trying to automate the importing of a Radarr/Sonarr SQLite database into Postgres 17 with pgloader v3.6.9 on alpine edge.

```
/app $ pgloader -V
pgloader version "3.6.9"
compiled with SBCL 2.3.6
/app $
```

I'm running:

```
pgloader --with "quote identifiers" --with "data only" /config/radarr.db "postgresql://radarr:radarr@localhost/radarr_main"
```

But I see the following failure:

```
KABOOM!
UNDEFINED-COLUMN: Database error 42703: column ""Id"" of relation "MovieFiles" does not exist
CONTEXT: PL/pgSQL function inline_code_block line 6 at FOR over SELECT rows
QUERY:
DO $$
DECLARE
  n integer := 0;
  r record;
BEGIN
  FOR r in
       SELECT 'select '
               || trim(trailing ')'
                  from replace(pg_get_expr(d.adbin, d.adrelid),
                               'nextval', 'setval'))
               || ', (select greatest(max(' || quote_ident(a.attname) || '), (select seqmin from pg_sequence where seqrelid = ('''
               || pg_get_serial_sequence(quote_ident(nspname) || '.' || quote_ident(relname), quote_ident(a.attname)) || ''')::regclass limit 1), 1) from only '
               || quote_ident(nspname) || '.' || quote_ident(relname) || '));' as sql
         FROM pg_class c
              JOIN pg_namespace n on n.oid = c.relnamespace
              JOIN pg_attribute a on a.attrelid = c.oid
              JOIN pg_attrdef d on d.adrelid = a.attrelid
                                 and d.adnum = a.attnum
                                 and a.atthasdef
        WHERE relkind = 'r' and a.attnum > 0
              and pg_get_expr(d.adbin, d.adrelid) ~ '^nextval'
              and c.oid in (select oid from reloids)
  LOOP
    n := n + 1;
    EXECUTE r.sql;
  END LOOP;

  PERFORM pg_notify('seqs', n::text);
END;
$$;
```
I have confirmed that the column `Id` **does** exist in the `MovieFiles` table

I notice in the output that the column name (`""Id"" `) is double-quoted - could this be an error introduced by using `--with "quote identifiers" ` ?

**Without** `--with "quote identifiers"`, I only get this far:

```
[...]

PR #1701, opened 2026-05-06
Fix error with quote identifiers and uppercase identifiers

til:
Fix fatal error when running with "quote identifiers" and the source has uppercase identifiers, as described in #1651. This PR simply contains the suggested fix from https://github.com/dimitri/pgloader/issues/1673#issuecomment-2930754447

Works for me, both with and without "quote identifiers", but I'm not familiar enough with pgloader to tell if it could have unintended consequences or if there is a better way.

### DIFF
--- a/src/pgsql/pgsql-create-schema.lisp
+++ b/src/pgsql/pgsql-create-schema.lisp
@@ -409,7 +409,7 @@ BEGIN
                   from replace(pg_get_expr(d.adbin, d.adrelid),
                                'nextval', 'setval'))
                || ', (select greatest(max(' || quote_ident(a.attname) || '), (select seqmin from pg_sequence where seqrelid = ('''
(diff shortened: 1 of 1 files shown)


dimitri:
Thanks for this fix, @til — you correctly identified that `pg_get_serial_sequence` expects a plain column name, not a quoted SQL identifier, so wrapping `a.attname` in `quote_ident()` broke sequence resets for tables with mixed-case column names when `quote identifiers` was active.

The same fix (removing the spurious `quote_ident(a.attname)` from the `pg_get_serial_sequence` call) was independently applied in commit [9cf7a2a](https://github.com/dimitri/pgloader/commit/9cf7a2a3e8371a1966f5c3b11ce55275211be0c7) as part of [#1705](https://github.com/dimitri/pgloader/pull/1705), so main already has the corrected behaviour. Closing as superseded — the bug is fixed.

PR #1704, opened 2026-06-15
Add verify-full to supported SSL modes for postgres

ragne:
I wanted to use the tool with databricks lakebase and it seems like pgloader doesn't support SNI with any of the existing `sslmode` values. 
Underlying `Postmodern` library [seem to support](https://marijnhaverbeke.nl/postmodern/#b5bb7222-8134-4dcb-83b7-f764b7f2bb33) SNI with mode set to `:full` and it looks like in the pgloader code only the parsing bits need to be fixed.

The only problem I see is that semantics are different from what `libpq` does. 
libpq sends SNI by default through the sslsni connection parameter, independently of the selected sslmode. In libpq, sslmode=require still uses TLS and sends SNI by default; sslmode=verify-full additionally verifies the server certificate hostname.
But cl-postgres from Postmodern [appears to pass](https://github.com/marijnh/Postmodern/blob/master/cl-postgres/protocol.lisp#L231) the hostname to the TLS layer only for the :full SSL mode.

Please take a look and let me know your thoughts.

### DIFF
--- a/docs/command.rst
+++ b/docs/command.rst
@@ -245,7 +245,7 @@ Where:
     *user* and *password*.
 
     The *sslmode* parameter values can be one of `disable`, `allow`,
-    `prefer` or `require`.
+    `prefer`, `require` or `verify-full`.
 
     For backward compatibility reasons, it's possible to specify the
     *tablename* option directly, without spelling out the `tablename=`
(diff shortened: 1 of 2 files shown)


dimitri:
Thanks for the fix, @ragne — clean and correct.

Rebased onto current main and opened [#1740](https://github.com/dimitri/pgloader/pull/1740) with your authorship preserved. Closing this one in favour of that PR.

PR #1719, opened 2026-06-23
Add default date format option for file loads

### LINKED ISSUE #1191: Idea: Specify a Date Format for all dates using a "With" option?
This is more an idea than a bug (is this the right place?):

I have a data file with many dates and having to add the column names just to set all the date formats to the same format seems rather _clunky_, so ...

pgloader currently allows specifying a Date Format on a column-by-column basis, but normally I would suspect that multiple dates would all follow the same format.

    Field Specification:
        column-name [date format 'YYYY-MM-DD HH24-MI-SS.US']

Idea:
Allow a catch-all Date Format to be specified using the "with" keyword, eg something like:

       WITH date format 'YYYY-MM-DD HH24-MI-SS.US'

Ideally, any column specific Field Specification Date Formats would then override the "WITH" generic one.

     Regards

sawirricardo:
## Summary
- accept `WITH date format ...` for CSV and fixed-width file loads
- apply the default format to target date/time columns while preserving per-field overrides
- document the option and update date-format parser/regression coverage

Fixes #1191.

## Verification
- `git diff --check`
- `docker run --rm -e USER=pgloader -v "$PWD/clojure":/work -w /work clojure:temurin-21-tools-deps clojure -M:test -m cognitect.test-runner -n pgloader.load-file.parser-test`
- SBCL reader check for edited Lisp source/parser files
- Clojure reader check for edited Clojure parser files


dimitri:
Thanks for this contribution — the design is right and the v3 implementation looks complete. I went through the diff carefully and found a few things that need fixing before we can merge.

**Bug 1 — v4 `fixed-option` grammar not updated (blocking)**

`WITH date format` is documented for `LOAD FIXED` and works on the v3 side, but in `clojure/src/pgloader/load_file/grammar.clj` the `fixed-option` rule was not updated. Only `csv-option` got `date-format` added. As it stands, `LOAD FIXED … WITH date format 'X'` will be a parse error in v4. One-line fix:

```
# grammar.clj line ~117 — add  | date-format  here:
fixed-option = truncate | create-tables | ... | fixed-header | date-format
```

**Bug 2 — positional index fallback in `ast.clj` can misapply formats (moderate)**

In `default-col-formats`, when a source column name is not found in `typed-by-name` (the name→type map of date/time target columns), the code falls back to the target projection at the same position index:

```clojure
(let [target-type (or (get typed-by-name col-name)
                      (:target-type (nth target-projections i nil)))]
```

If the source and target column lists are in a different order, this picks the wrong type and can apply a date format to a non-date column. Since `typed-by-name` already captures all date/time columns by name, the index fallback isn't needed — dropping `(nth target-projections i nil)` and just using `(get typed-by-name col-name)` is sufficient and safe.

**Nit — silent no-op**

If `WITH date format` is specified but no `TARGET TABLE (col date)` type annotations are present, the option is silently ignored. A `log/warn` here would save debugging time for users who hit this.

---

Overall the approach is solid and the v3 side is correct. Would you like to submit a revised version with those two fixes applied? Happy to merge once CI is green on the updated branch.

sawirricardo:
would you mind to recheck again @dimitri ?

dimitri:
Thanks again for this contribution — the implementation is solid and all three review points were addressed cleanly.

I've opened PR #1732 as a follow-up that cherry-picks both your commits verbatim (authorship preserved) and adds one additional fix on top: in v4, the LOAD FIXED path accepted `WITH date format` in the grammar and stored it in `with-options`, but never populated `source.column-formats`, so the format was silently ignored at runtime. PR #1732 fills that gap by mirroring the CSV path's `default-col-formats` logic in the FIXED transform and wiring `apply-date-format` into `fixed.clj`'s `read-rows`.

Closing this PR in favour of #1732 which is rebased on current main.

Issue #1676, opened 2025-06-23
Invalid docs for default MYSql to Postgres migration at Casting rules


FluffyDiscord:
Hi, I need to change the default datetime cast rules. However simply copying them to `.load` fails with parsing error.

My load
```
LOAD DATABASE
	FROM mysql://db:db@127.0.0.1:12826/db
	INTO postgresql://db:db@127.0.0.1:5433/db
WITH
    create indexes, uniquify index names, foreign keys,
    include drop,
    reset sequences, multiple readers per thread,
    workers = 8, concurrency = 2, rows per range = 50000
 
SET PostgreSQL PARAMETERS
    maintenance_work_mem to '1GB'
 
SET MySQL PARAMETERS
    net_read_timeout  = '31536000',
    net_write_timeout = '31536000',
    lock_wait_timeout = '31536000'

CAST
    type datetime when default "0000-00-00 00:00:00" and not null
      to timestamp drop not null drop default
          using zero-dates-to-null,
    
    type datetime when default "0000-00-00 00:00:00"
      to timestamp drop default
          using zero-dates-to-null,
    
    type datetime with extra on update current timestamp when not null
      to timestamp drop not null drop default
         using zero-dates-to-null,
    
    type datetime with extra on update current timestamp
      to timestamp drop default
         using zero-dates-to-null,
    
    type timestamp when default "0000-00-00 00:00:00" and not null
      to timestamp drop not null drop default
          using zero-dates-to-null,
    
    type timestamp when default "0000-00-00 00:00:00"
      to timestamp drop default
          using zero-dates-to-null,
    
    type date when default "0000-00-00" to date drop default
          using zero-dates-to-null,

    type datetime to timestamp,
    type timestamp to timestamp
;
```

the error:
```
What I am doing here?

At

  CAST
      type datetime when default "0000-00-00 00:00:00" and not null
  
                                                      ^ (Line 19, Column 52, Position 532)

In context COMMAND:

While parsing COMMAND. Expected:

     the character Tab
  or the character Newline
  or the character Return
  or the character Space
  or the character , (COMMA)
  or the string "--"
  or the string "/*"
  or the character ; (SEMICOLON)
  or the string "after"
  or the string "alter"
  or the string "before"
  or the string "cast"
  or the string "decoding"
  or the string "distribute"
  or the string "drop"
  or the string "excluding"
  or the string "including"
  or the string "keep"
  or the string "materialize"
  or the string "set"
  or the string "to"
[...]

Issue #1670, opened 2025-05-15
How to support  DBF Chinese Big5 code (CP950)?


raindy29:
I use pgloader -E but can't see cp950 code, Chinese characters are not correct.

Issue #1187, opened 2020-07-11
SQLite and preserving casing


deltaGPhys:
I've seen [this issue](https://github.com/dimitri/pgloader/issues/1105), but get the error below when I run:

`pgloader --with "quote identifiers" /Users/..../db.sqlite3 postgresql:///dbname`

```
#<CLOSURE (LAMBDA (COMMAND-LINE-ARGUMENTS::VALUE)
            :IN
            COMMAND-LINE-ARGUMENTS::MAKE-OPTION-ACTION) {1004DBA85B}> fell through ETYPECASE expression.
Wanted one of (NULL KEYWORD SYMBOL CONS).
pgloader [ option ... ] command-file ...
pgloader [ option ... ] SOURCE TARGET
  --help -h                       boolean  Show usage and exit.
  --version -V                    boolean  Displays pgloader version and exit.
  --quiet -q                      boolean  Be quiet
```

PR #1721, opened 2026-06-23
test: coverage for Fix A–D, F — MySQL cast, DBF charset, SQLite CLI --with

dimitri:
## Summary

Adds targeted test coverage for the fixes in [#1720](https://github.com/dimitri/pgloader/pull/1720). All unit tests pass locally (158 tests, 582 assertions, 0 failures).

---

### MySQL CAST `and not null` — integration test (#1676)

**What**: new table `zero_dates_notnull` in `mytest.sql` with:
- `created_at DATETIME NOT NULL DEFAULT '0000-00-00 00:00:00'` — matched by the `and not null` cast rule → NOT NULL dropped → nullable timestamp in PG
- `updated_at DATETIME DEFAULT '0000-00-00 00:00:00'` — NOT matched (nullable) → fallback rule → nullable timestamp in PG

Two new SQL checks (sql/13, sql/14): column nullability from information_schema and COUNT of zero-date NULLs. The `and not null` guard is verified by the fact that `created_at` becomes nullable (via `drop not null`) while `updated_at` needed no such action.

**v3 and v4**: same expected output since both now parse the guard identically.

---

### DBF charset alias — unit test (#1670)

**What**: `clojure/test/pgloader/source/dbf_test.clj` — verifies that `cp950`/`CP950` → `Big5` and `cp932`/`CP932` → `windows-31j` resolve to valid `Charset` objects via the JVM, covering the alias normalisation map added in v4.

---

### SQLite CLI `--with` — grammar + unit + integration tests (#1187)

**Grammar (v4)**: 3 tests in `parser_test.clj` verify that `LOAD DATABASE FROM sqlite://... WITH quote/snake_case/downcase identifiers` parses and sets the correct `with-options` key.

**CLI (v4)**: test in `cli_test.clj` that `--with 'quote identifiers'` for a SQLite URI is captured in `:with-opts` without error.

**Integration (v3 + v4)**: new `sqlite-cli-with` test target in the SQLite Makefile. Invokes pgloader via URI args (`sqlite:///...  pgsql://...`) with `--with 'quote identifiers'`; verifies `CamelTable` appears with its exact mixed-case name in PostgreSQL and all 2 rows are present. This exercises the v3 fix (ecase → guarded case in `parse-cli-options`) end-to-end in CI.


dimitri:
Folding this coverage back into #1720 — no value as a separate PR.

PR #1759, opened 2026-07-15
ci: pin sigstore/cosign-installer@main to a full commit SHA

kobihikri:
Hi, and thank you for pgloader.

Small supply-chain hardening. Both `.github/workflows/docker-publish.yml` (line 42) and `.github/workflows/docker-publish-v4.yml` (line 30) use the third-party action `sigstore/cosign-installer@main`. `@main` is a mutable branch, and in both files the `build` job declares `permissions: packages: write` + `id-token: write` and logs in to ghcr.io with `GITHUB_TOKEN`. If that upstream branch were repointed or compromised (the `tj-actions/changed-files` class of risk), it would run with the job's write-scoped token and OIDC identity on push/tag builds.

This PR pins the action to a full commit SHA (`v4.1.2`). The installed cosign version is unchanged — the `cosign-release: v2.2.3` input is untouched — so this only removes the mutable reference. The other actions in these jobs are already version-pinned.

For transparency: I used AI assistance to help identify this and draft the change; I verified every line against the live workflows myself.


dimitri:
Thanks for the careful writeup and for verifying every line yourself — the risk framing (OIDC + packages write on a `@main`-pinned action) is spot-on. Merging.

PR #1762, opened 2026-07-21
Fix typo in docs

jelder:
The docs had "from migration from" which didn't make grammatical sense in context. I think the author may have intended to say "live migration from."


dimitri:
Thanks @jelder !

